PR #11, opened 2026-08-12
Handle Expo dev server disconnections

krystofwoldrich:
## Summary

- send an immediate device-list heartbeat and continue broadcasting heartbeats every two seconds while dashboards are subscribed
- require valid Hub protocol messages before marking the browser connected, detect silent connections after six seconds, and reconnect with bounded exponential backoff
- replace the dashboard with a blocking Expo-styled status page while connecting or disconnected, with guidance to restart the dev server and reload

## Why

The device-list WebSocket previously emitted only when the discovered device list changed. The browser retried explicit socket closes, but it exposed no connection state and could not detect a half-open connection, so a closed dev server could leave stale interactive UI visible.

## User impact

Users now see a clear full-page connecting or disconnected state instead of a stale dashboard. Device streaming is stopped while the server is unavailable, and the page provides a direct reload action after the user restarts the Expo dev server in their terminal.

## Validation

- bun install
- bun run test
- bun run lint
- bun run build

### DIFF
--- a/packages/@expo/hub-components/src/components/icons.tsx
+++ b/packages/@expo/hub-components/src/components/icons.tsx
@@ -153,6 +153,40 @@ export function PowerIcon({ size = 16, color = 'currentColor', strokeWidth = 1.6
   );
 }
 
+/** Lucide-inspired `unplug` — two separated cable ends for connection-loss states. */
+export function CableDisconnectIcon({
+  size = 16,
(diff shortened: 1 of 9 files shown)


krystofwoldrich-agent:
Superseded by #20, which rebases this work onto the latest upstream main and preserves the newer dashboard changes.

PR #37, opened 2026-09-01
Improve agent interaction cursor UI

krystofwoldrich-agent:
## Summary

- render agent interactions with a named pink cursor and matching device status
- replace the blocking takeover overlay with a viewer cursor, attached `you` label, and compact interruption note
- keep cursor labels visible beyond device frames and adapt their attached corners to available viewport space
- keep warning dismissal scoped to the continuous agent-active session and restore the native cursor after inactivity
- share cursor geometry and use semantic color, radius, border, and typography tokens

## Test plan

- `bun test` focused interaction, animation, layout, activity, and device-status suites — 26 passing
- `bun run --cwd packages/@expo/hub-client typecheck`
- `bun run --cwd packages/@expo/hub-components typecheck`
- `bun run --cwd packages/@expo/hub-client lint`
- `bun run --cwd packages/@expo/hub-components lint`
- `bun run build`
- validated the hover, user-interaction, subsequent-agent-action, edge-placement, overflow, and inactivity transitions with real Argent interactions in Safari through the running device hub

### DIFF
--- a/packages/@expo/hub-client/package.json
+++ b/packages/@expo/hub-client/package.json
@@ -37,14 +37,17 @@
     "format:check": "oxfmt --check"
   },
   "peerDependencies": {
-    "react": "*"
+    "react": "*",
+    "react-dom": "*"
   },
   "devDependencies": {
     "@swc/cli": "^0.8.1",
     "@swc/core": "^1.15.43",
     "@types/bun": "latest",
     "@types/react": "~19.2.2",
+    "@types/react-dom": "~19.2.0",
     "oxfmt": "^0.56.0",
+    "react-dom": "catalog:",
     "oxlint": "^1.71.0",
     "typescript": "~6.0.3"
   },
--- a/packages/@expo/hub-client/src/DeviceScreen.tsx
+++ b/packages/@expo/hub-client/src/DeviceScreen.tsx
@@ -7,6 +7,7 @@ import {
   useRef,
   useState,
 } from 'react';
+import { createPortal } from 'react-dom';
 
 import { streamGeometry } from './orientation';
 import { AgentInteractionIndicator } from './AgentInteractionIndicator';
@@ -107,6 +108,7 @@ export function DeviceScreen({
   borderRadius,
   squircle,
   agentInteraction,
+  agentInteractionPortalTarget,
 }: DeviceScreenProps) {
   const { videoKind, attachVideo, sendTouch, sendMultiTouch, sendKey, screen, status, error } =
     client;
@@ -378,7 +380,14 @@ export function DeviceScreen({
         </>
       )}
 
-      <AgentInteractionIndicator interaction={agentInteraction} />
+      {agentInteractionPortalTarget === undefined ? (
+        <AgentInteractionIndicator interaction={agentInteraction} />
+      ) : agentInteractionPortalTarget ? (
+        createPortal(
+          <AgentInteractionIndicator interaction={agentInteraction} />,
+          agentInteractionPortalTarget
+        )
+      ) : null}
 
       {status !== 'streaming' && (
         <div
--- a/packages/@expo/hub-components/package.json
+++ b/packages/@expo/hub-components/package.json
@@ -40,6 +40,7 @@
     "format:check": "oxfmt --check"
   },
   "dependencies": {
+    "@expo/hub-client": "workspace:*",
     "@expo/styleguide": "^14.1.5",
     "@radix-ui/react-dialog": "^1.1.15",
     "@radix-ui/react-dropdown-menu": "^2.1.16",
@@ -51,7 +52,6 @@
     "react-dom": "*"
   },
   "devDependencies": {
-    "@expo/hub-client": "workspace:*",
     "@swc/cli": "^0.8.1",
     "@swc/core": "^1.15.43",
     "@types/react": "~19.2.2",
--- a/packages/@expo/hub-components/src/__tests__/agent-device-status.test.tsx
+++ b/packages/@expo/hub-components/src/__tests__/agent-device-status.test.tsx
@@ -1,11 +1,12 @@
 import { describe, expect, test } from 'bun:test';
 import { readFileSync } from 'node:fs';
 import { type DeviceClient } from '@expo/hub-client';
+import { devicePointerLabelRadius } from '@expo/hub-client';
 import { renderToStaticMarkup } from 'react-dom/server';
 
 import { DeviceListItem } from '../components/DeviceListItem';
 import { shouldCompactAgentDeviceStatus } from '../components/useCompactAgentDeviceStatus';
-import { AgentDeviceOverlay } from '../dashboard/AgentDeviceOverlay';
+import { AgentDeviceOverlay, agentDeviceOverlayPlacement } from '../dashboard/AgentDeviceOverlay';
 import { type Device } from '../dashboard/data';
 import { type DeviceFrameAssets } from '../dashboard/deviceFrame';
 import { PhoneFrame } from '../dashboard/PhoneFrame';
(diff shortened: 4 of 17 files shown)


packages/@expo/hub-components/package.json
```diff
@@ -40,6 +40,7 @@
     "format:check": "oxfmt --check"
   },
   "dependencies": {
+    "@expo/hub-client": "workspace:*",
```
krystofwoldrich-agent, line 43: **[design]** `@expo/hub-client` moves from `devDependencies` to `dependencies`, and `AgentDeviceOverlay.tsx:9` / `PhoneFrame.tsx:111` import runtime values from it. `src/index.ts:48-51` documents the opposite contract: `DeviceScreen` is injected "so this library never imports the client at runtime", and the package description calls this a "Dependency-free UI kit shared by Expo Hub and the Expo dashboard website". The website consumer would now pull the device-client package into its bundle for a few style constants and one hook.

Options: keep the pointer presentation in `hub-components` and pass it to the injected `DeviceScreen` through props (the existing injection pattern), or let the consumer (`expo-device-hub`, which already depends on both) inject the overlay. If the boundary change is intentional, update the `index.ts` comment and the package description in this PR.

packages/@expo/hub-components/src/theme/tokens.ts
```diff
@@ -114,7 +123,12 @@ export const textSize: Record<'2xl' | 'xl' | 'lg' | 'base' | 'sm' | 'xs' | '2xs'
   lg: { fontSize: 18, fontWeight: 400, lineHeight: 1.4, letterSpacing: '0' },
   base: { fontSize: 16, fontWeight: 400, lineHeight: 1.6, letterSpacing: '0' },
   sm: { fontSize: 14, fontWeight: 400, lineHeight: 1.6, letterSpacing: '0' },
-  xs: { fontSize: 12, fontWeight: 400, lineHeight: 1.6, letterSpacing: '0' },
+  xs: {
```
krystofwoldrich-agent, line 126: **[design]** `textSize` is a public export. `index.ts:9-16` says inline-style components work in any environment that defines the `--expo-theme-*` variables and that the CSS variables are not bundled. A consumer that supplies its own variables will not define `--expo-text-size-xs-*`, so every `textSize.xs` usage (Button `2xs`/`xs`, Select, SidebarRow, LogRow, ...) gets an unresolved `font-size`. The change also widens `TextStyle` to `number | string` for all sizes while only `xs` uses strings.

The motivation (hub-client cannot import hub-components) does not require this: `device-pointer-presentation.ts` already hardcodes `height: 29` and `padding: '0 12px'`, so `fontSize: 12, lineHeight: 1.6` there keeps parity without changing the public token shape.

packages/@expo/hub-client/package.json
```diff
@@ -37,14 +37,17 @@
...
+    "react-dom": "*"
   },
   "devDependencies": {
     "@swc/cli": "^0.8.1",
     "@swc/core": "^1.15.43",
     "@types/bun": "latest",
     "@types/react": "~19.2.2",
+    "@types/react-dom": "~19.2.0",
     "oxfmt": "^0.56.0",
+    "react-dom": "catalog:",
```
krystofwoldrich-agent, line 50: **[design]** `bun.lock` is not updated for the new `react-dom` / `@types/react-dom` entries here, nor for `@expo/hub-client` moving to `dependencies` in hub-components. `bun install --frozen-lockfile` fails at the PR head with "lockfile had changes, but lockfile is frozen", and `.github/workflows/release.yml:42` uses that flag. Run `bun install` and commit the lockfile.

krystofwoldrich-agent (commented):

**Verdict:** REQUEST_CHANGES
**Findings:** 0 critical · 4 design · 2 suggestion · 1 nit

The PR replaces the blocking "Take over anyway" overlay with a pointer-following "you" cursor, gives the agent cursor a pink label, portals the agent indicator outside the clipped screen so labels stay visible over the device bezel, and introduces `agent.*` color tokens. The UI logic is sound and the tests cover it. The main concern is a package-boundary change: `@expo/hub-components` now imports `@expo/hub-client` at runtime, which reverses the boundary documented in its own `index.ts` (lines 48-51) and `package.json` description. The PR also does not update `bun.lock`, and it no longer merges cleanly with `main` (conflicts in `packages/@expo/hub-client/package.json` and `agent-device-status.test.tsx`).

Checks run at the PR head: hub-client `bun test` 109 pass; hub-components `bun test` 19 pass; `tsc --noEmit` and `oxlint` pass in both packages. `bun install --frozen-lockfile` fails (lockfile not updated). `oxfmt --check` fails repo-wide with no config; this is pre-existing.

**Findings**
- [design] Runtime dependency on `@expo/hub-client` reverses the documented package boundary — `packages/@expo/hub-components/package.json:43`
- [design] `textSize.xs` becomes CSS variables that only `theme.css` defines — `packages/@expo/hub-components/src/theme/tokens.ts:126`
- [design] `bun.lock` not updated; `bun install --frozen-lockfile` fails — `packages/@expo/hub-client/package.json:50`
- [design] Does not merge cleanly with `main`; trailing-comma reformatting of untouched lines caused the test-file conflict — `packages/@expo/hub-components/src/__tests__/agent-device-status.test.tsx:184`
- [suggestion] Callout placement measures the viewport, but the stream card (`StreamPanel.tsx:94`, `overflow: hidden`) is the real clip — `packages/@expo/hub-components/src/dashboard/PhoneFrame.tsx:126`
- [suggestion] `undefined` vs `null` carry different meanings for `agentInteractionPortalTarget` — `packages/@expo/hub-client/src/DeviceScreen.tsx:383`
- [nit] `maxWidth` on the interruption warning has no effect with `nowrap` children — `packages/@expo/hub-components/src/dashboard/AgentDeviceOverlay.tsx:172`

packages/@expo/hub-components/src/__tests__/agent-device-status.test.tsx
```diff
@@ -100,67 +103,147 @@ describe('agent device status', () => {
...
 
-  test('keeps the agent takeover button borderless in light mode', () => {
+  test('leaves the normal device cursor visible when no agent is interacting', () => {
+    const markup = renderToStaticMarkup(
+      <PhoneFrame
+        device={{ ...IPHONE, deviceFrame: null }}
+        client={STREAMING_CLIENT}
+        DeviceScreen={() => <div role="application" style={{ cursor: 'crosshair' }} />}
+        displayScreen={() => null}
+      />,
```
krystofwoldrich-agent, line 184: **[design]** This PR no longer merges cleanly with `main`: `git merge-tree` reports conflicts in this file and in `packages/@expo/hub-client/package.json`. The conflict here comes mostly from adding trailing commas after multi-line JSX arguments (`/>,`) on lines the PR does not otherwise change; that style appears nowhere else in the repo (`git grep` on `main` finds zero matches), and `main` edited the same regions (`pixel-9` frame asset, `device-frame-anchor` regex). A rebase is needed; keeping the surrounding formatting would avoid similar conflicts.

packages/@expo/hub-components/src/dashboard/PhoneFrame.tsx
```diff
@@ -97,30 +108,35 @@ export function PhoneFrame({
...
+  }, [activeAgentInteraction]);
+
+  const positionPointerCallout = (event: ReactPointerEvent<HTMLDivElement>) => {
+    const frame = event.currentTarget.getBoundingClientRect();
+    pointerOverlayRef.current?.position({
+      clientX: event.clientX,
+      clientY: event.clientY,
+      frameLeft: frame.left,
+      frameTop: frame.top,
+      viewportWidth: window.innerWidth,
```
krystofwoldrich-agent, line 126: **[suggestion]** `agentDeviceOverlayPlacement` flips the callout when the browser viewport runs out of room, but the frame's ancestor in `StreamPanel.tsx:94` has `overflow: 'hidden'`. The stream card is the real clipping boundary and is narrower than the viewport whenever a sidebar is open, so the callout can pass the viewport test and still be cut off. Measure against the clipping container's rect instead (or pass it in from `StreamPanel`).

packages/@expo/hub-client/src/DeviceScreen.tsx
```diff
@@ -378,7 +380,14 @@ export function DeviceScreen({
         </>
       )}
 
-      <AgentInteractionIndicator interaction={agentInteraction} />
+      {agentInteractionPortalTarget === undefined ? (
```
krystofwoldrich-agent, line 383: **[suggestion]** `undefined` renders inline, `null` renders nothing, and an element portals. This is a public prop on `DeviceScreenProps` (`types.ts:426`); a consumer that passes `someRef.current` before mount silently loses the indicator. A simpler contract is "falsy → inline, element → portal", which also removes the extra empty render `PhoneFrame` does before its ref callback sets state.

packages/@expo/hub-components/src/dashboard/AgentDeviceOverlay.tsx
```diff
@@ -1,61 +1,194 @@
...
+            you
+          </span>
+          {warningVisible && (
+            <span
+              role="note"
+              data-testid="agent-interruption-warning"
+              style={{
+                display: 'block',
+                width: 'max-content',
+                maxWidth: `calc(100vw - ${CALLOUT_MARGIN * 2}px)`,
```
krystofwoldrich-agent, line 172: **[nit]** Both child lines below are `whiteSpace: 'nowrap'` (lines 182 and 185), so this `calc(100vw - 32px)` cap never wraps the text. Either drop `maxWidth` or drop `nowrap` on the children.

PR #31, opened 2026-08-28
Add WebRTC stream statistics

krystofwoldrich-agent:
## ⛔ Blocked — do not merge

This PR depends on https://github.com/expo/serve-emu/pull/15, which moves the Android font-scale, network, and WebRTC statistics endpoints into the `expo/serve-emu` package. Do not merge this PR until that dependency lands and this PR points at its merged commit.

All `/vendor/serve-emu/**` routes in this branch are now transparent package routes; Expo Device Hub no longer injects or intercepts those endpoints.

## Summary

- collect one-second WebRTC receiver telemetry for browser-presented FPS, received bitrate, loss, jitter, jitter-buffer delay, dropped frames, freezes, RTT, and ICE path
- consume viewer-scoped sender/capture telemetry from serve-sim on iOS and the package-owned `/webrtc/stats` endpoint from serve-emu on Android
- derive Android Publisher FPS and H.264 payload bitrate from cumulative, viewer-scoped publisher counters
- show Stream, Client, Encoder, and Capture groups in a compact responsive two-column layout with Client FPS and bitrate graphs
- omit structurally unavailable Android fields instead of displaying permanent dashes

## Telemetry semantics

Client FPS counts frames actually presented through `requestVideoFrameCallback`; client bitrate uses inbound RTP `bytesReceived`. Loss, drops, freezes, and jitter-buffer delay are calculated over each real sample window, while jitter, RTT, and ICE path are current browser receiver values.

On Android, Publisher submissions are H.264 frames accepted by the native media track, Publisher drops are keyframe-gate/backpressure drops rather than packet loss, and Payload bitrate excludes RTP/SRTP/UDP overhead and retransmissions. Source/output values describe the shared scrcpy encoder; session counters are selected by the browser WebRTC session UUID.

## Test plan

- `bun run test` — 142 Hub tests pass
- `bun run typecheck`
- `bun run lint`
- `bun run build`
- serve-emu: 665 tests, server/UI/test typechecks, and production build pass
- launched the built standalone CLI with `--platform android --transport webrtc --video-fps 30 --video-bitrate 8000000`
- used agent-browser input to open Chrome and swipe its content; `/api/foreground` reported `com.android.chrome`
- confirmed `/webrtc/stats` selected the exact browser UUID, reported `connected: true` and a direct host/prflx UDP path, and increased source/submitted frames by 18 with 303,515 additional H.264 payload bytes and zero publisher drops during one swipe
[...]

PR #39, opened 2026-09-02
Add WebRTC stream statistics

krystofwoldrich:
This replaces #31 with the same telemetry layer rebased onto current `main` and published from the Expo repository so the follow-up resolution and MMAP work can form a native GitHub stack.

The serve-emu dependency is supplied by merged Device Hub PR #48 at serve-emu commit `2b161b5a39940716eca45262dee6fbc3c229979c`.

## Summary

- collect one-second WebRTC receiver telemetry for presented FPS, received bitrate, loss, jitter, jitter-buffer delay, dropped frames, freezes, RTT, and ICE path
- consume package-owned Android stream statistics and viewer-scoped publisher counters
- show Stream, Client, Encoder, and Capture groups with FPS and bitrate graphs
- keep all `/vendor/serve-emu/**` routes transparent

## Test plan

- [x] `bun install --frozen-lockfile`
- [x] `bun run test`
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run build`
- [x] Integrated browser: Android WebRTC stream reached Live, telemetry updated after device interaction, and browser warnings/errors remained empty


krystofwoldrich:
Local validation after the rebase/migration: `bun run test` and `bun run typecheck` both pass on this exact branch.

packages/@expo/hub-client/src/stream-stats.ts
```diff
@@ -0,0 +1,534 @@
...
+    samples: [],
+    encoder: null,
+    capture: null,
+    stale: false,
+    serverStale: false,
+  };
+}
+
+/** Own the single stats poll for a WebRTC peer so UI remounts do not reset history. */
+export function useWebRtcStreamStats(
```
krystofwoldrich, line 401: **[suggestion]** `useWebRtcStreamStats` starts polling as soon as a peer connection exists: one `getStats()` and one `/webrtc/stats` fetch per second per viewer, plus a second interval for staleness. The Stream options section is collapsed by default (`inspectorSections.test.tsx` asserts three collapsed sections), so most viewers pay this cost without seeing the data.

Consider an `enabled` input driven by the section's open state, and reset the history when disabled. If always-on history is intentional so the graph is already full when the panel opens, a short comment here would stop the next reader from "optimizing" it away.

packages/@expo/hub-components/src/dashboard/StreamOptionsSection.tsx
```diff
@@ -101,6 +119,457 @@ function withCurrentValue(
...
+          whiteSpace: 'nowrap',
+        }}
+      >
+        {value}
+      </span>
+    </div>
+  );
+}
+
+function StreamStatistics({
```
krystofwoldrich: **[suggestion]** This PR extracts `MetricChart` into its own file, which is the right pattern. `StreamStatistics` and its ten formatters land inline instead and grow this file from ~320 to ~800 lines (#41 adds ~100 more). A `StreamStatistics.tsx` next to `MetricChart.tsx` would keep `StreamOptionsSection` focused on controls and make the statistics table independently testable.

packages/@expo/hub-client/src/stream-stats.ts
```diff
@@ -0,0 +1,534 @@
...
+  };
+}
+
+function readServeEmuEncoderStats(
+  source: Record<string, unknown>,
+  session: Record<string, unknown> | null,
+): DeviceStreamEncoderStats {
+  return {
+    codec: nonEmptyString(source.codec),
+    encodeFps: finiteNumber(source.fps),
```
krystofwoldrich, line 287: **[nit]** For serve-emu, `encodeFps` here and `serverFps` (line 327) both read `source.fps`. The Android UI shows the same number twice, as "Server FPS" and "Output FPS". Either drop the Android "Output FPS" row or leave `encodeFps` null here so the row hides, matching how the other unavailable serve-emu values are handled.

packages/expo-device-hub/package.json
```diff
@@ -43,6 +43,7 @@
     "/expo-module.config.json"
   ],
   "dependencies": {
+    "@fastify/busboy": "3.2.0",
```
krystofwoldrich: **[nit]** This is needed because the vendored serve-emu middleware imports `@fastify/busboy` at runtime (serve-emu's `multipart-upload.ts`) and `scripts/vendor.ts` strips the vendored package.json. A one-line note next to `vendorDependencies` or in the PR body would save the next person a search. This is also the only exact pin among the runtime dependencies; serve-emu itself declares `^3.2.0`.

.changeset/webrtc-stream-statistics.md
```diff
@@ -0,0 +1,5 @@
+---
+'expo-device-hub': minor
+---
+
+Add WebRTC stream statistics.
```
krystofwoldrich: **[nit]** The PR also deletes the Hub's `/api/network` and `/api/font-scale` compatibility handlers (`serve-emu-device-options.ts`) because serve-emu now serves them. A second sentence here would record that ownership move in the changelog.

krystofwoldrich (commented):

## Review: Add WebRTC stream statistics (PR 1/3 in stack: #39 → #40 → #41)

**Verdict:** COMMENT (no blocking findings)
**Findings:** 0 critical · 0 design · 2 suggestion · 3 nit

This PR adds a client-side WebRTC telemetry poller (`stream-stats.ts`), threads `streamStats` through the `DeviceClient` contract, and renders grouped Stream / Client / Encoder / Capture statistics in Stream options. It also deletes the Hub's `/api/network` and `/api/font-scale` compatibility handlers. I checked serve-emu at the pinned commit (`2b161b5`): its middleware serves both routes, so the deletion is safe. The counter-delta logic keeps "unknown" distinct from zero and is well covered by tests.

### Findings

- **[suggestion]** `packages/@expo/hub-client/src/stream-stats.ts:401` — stats polling runs for every WebRTC viewer, even with the panel collapsed. Inline.
- **[suggestion]** `packages/@expo/hub-components/src/dashboard/StreamOptionsSection.tsx:286` — `StreamStatistics` (~300 lines) belongs in its own file, like the `MetricChart` extraction in this PR. Inline.
- **[nit]** `packages/@expo/hub-client/src/stream-stats.ts:287` — Android "Server FPS" and "Output FPS" show the same value. Inline.
- **[nit]** `packages/expo-device-hub/package.json:46` — `@fastify/busboy` is an exact pin with no note on why it is needed. Inline.
- **[nit]** `.changeset/webrtc-stream-statistics.md:5` — changeset omits the compat-route removal. Inline.

**Stack note:** the npm `files` fix in #41 (`vendor/serve-emu/dist/shared/*.js`) is required for this PR too. The serve-emu commit this PR depends on already imports `./shared/api-contracts.js` from `dist/middleware.js`. See the #41 review for details.

---

packages/expo-device-hub/package.json
krystofwoldrich: Add an artificial field which lists vendor dependencies from serve-sim and serve-emu inside of the expo-device-hub package.json.

packages/@expo/hub-client/src/stream-stats.ts
krystofwoldrich: Fixed in `cdc8742`: WebRTC statistics polling now has an explicit enabled state driven by the Stream options section. Closing the section stops both polling intervals, aborts the server request, clears retained statistics, and resets the next measurement window.

packages/@expo/hub-client/src/stream-stats.ts
krystofwoldrich: Fixed in `cdc8742`: removed the Android `Output FPS` row and updated the Android telemetry test to ensure it is not rendered. `Server FPS` remains the single source-FPS reading.

.changeset/webrtc-stream-statistics.md
krystofwoldrich: Fixed in `cdc8742`: the changeset now records that Android network and font-scale compatibility routes moved to serve-emu.

packages/@expo/hub-components/src/dashboard/StreamOptionsSection.tsx
krystofwoldrich: Fixed in `cdc8742`: moved the statistics renderer and formatters into `StreamStatistics.tsx`. `StreamOptionsSection.tsx` is back to focusing on stream controls, and the paused-state test now renders `StreamStatistics` directly.

packages/expo-device-hub/package.json
krystofwoldrich: Fixed in `cdc8742`: added `vendorRuntimeDependencies` metadata grouped by `@expo/serve-sim` and `serve-emu`, and aligned `@fastify/busboy` with serve-emu's `^3.2.0` range.

PR #40, opened 2026-09-02
Add Android stream resolution and source controls

krystofwoldrich:
## Summary

- add runtime Android stream resolution selection while preserving the Hub session across encoder restarts
- add the scrcpy/WebRTC versus emulator gRPC source selector
- use the 60 FPS capture default and expose the pipeline diagnostics needed by the stream statistics UI
- correct FPS accounting so client FPS counts browser-presented frames

This PR contains only the Device Hub resolution/source-selection and FPS-accounting layer. It is based on #39 and deliberately leaves the serve-emu pointer at that lower layer dependency.

## Test plan

- [ ] `bun run test`
- [ ] `bun run typecheck`
- [ ] `bun run lint`
- [ ] `bun run build`


krystofwoldrich:
Local standalone validation: `bun run test` and `bun run typecheck` both pass on this exact layer.

packages/@expo/hub-client/src/useAndroidDevice.ts
```diff
@@ -397,6 +459,235 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
...
+    void refresh();
+    const timer = setInterval(refresh, STREAM_SOURCE_POLL_MS);
+    return () => {
+      cancelled = true;
+      clearInterval(timer);
+      controller?.abort();
+    };
+  }, [streamSourceUrl]);
+
+  const updateStreamSettings = useCallback(
```
krystofwoldrich: **[design]** `updateStreamSettings` is a near-verbatim copy of `useIosDevice.ts` lines 630–673: same refs, same optimistic write, same request-id guard, same `finally` block. The GET side below (line 628) mirrors `useIosDevice.ts` lines 1260–1296 and adds polling. The real differences are `androidStreamSettingsPatch` / `parseAndroidStreamSettings` and the poll interval.

A shared `useStreamSettingsResource({ url, parse, toPatch, pollMs })` hook in hub-client would let both platforms share one state machine and one test suite. #41 copies `setStreamSource` again for `setGrpcImageMode`, so the duplication compounds up the stack.

packages/@expo/hub-client/src/useAndroidDevice.ts
```diff
@@ -397,6 +459,235 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
...
+          if (!response.ok) throw new Error(`Stream source update failed (${response.status})`);
+          const next = parseAndroidStreamSource(await response.json());
+          if (!next) throw new Error('Stream source update returned an invalid response');
+          if (streamSourceRequestRef.current === request) {
+            streamSourceRef.current = next;
+            setStreamSourceState(next);
+          }
+        })
+        .catch(() => {
+          // The server stages source changes atomically, so the previous source
```
krystofwoldrich, line 490: **[suggestion]** A failed PUT is swallowed here. From the user's side the Source toggle greys out, then snaps back with no explanation. The likely real-world failure is the emulator gRPC endpoint not being reachable (a physical device, or an emulator started without the gRPC port). Consider surfacing the error text through the client's existing `error` channel or a `streamSourceError` field so the UI can say why the switch did not take.

packages/@expo/hub-client/src/useAndroidDevice.ts
```diff
@@ -135,6 +162,22 @@ export function androidWsUrlFor(
   return u.toString();
 }
 
+/** Count frames presented between browser callbacks, including skipped callbacks. */
+export function presentedVideoFrameDelta(
```
krystofwoldrich: **[suggestion]** Good change: counting `metadata.presentedFrames` deltas fixes undercounting when the browser skips callbacks. iOS still calls `markWebRtcFrameDecoded()` once per callback (`useIosDevice.ts:850`), so the "Client FPS" statistic added in #39 now means different things on the two platforms under load. Consider moving `presentedVideoFrameDelta` to a shared module and applying it in the iOS frame callback too.

packages/@expo/hub-client/src/useAndroidDevice.ts
```diff
@@ -62,6 +74,8 @@ const KEYFRAME_REQUEST_COOLDOWN_MS = 1500;
 const FOREGROUND_POLL_MS = 5000;
 const EVENTS_POLL_MS = 1000;
 const STREAM_METADATA_POLL_MS = 1500;
+const STREAM_SETTINGS_POLL_MS = 3000;
+const STREAM_SOURCE_POLL_MS = 3000;
```
krystofwoldrich: **[suggestion]** With this PR an active Android client runs eight independent intervals: events 1s, stream metadata 1.5s, device settings 3s, stream settings 3s, stream source 3s, foreground 5s, plus the two 1s stats timers from #39. Stream settings and stream source change only after a user action and both come from the same serve-emu router. Options: poll both in one loop, back off while `document.hidden`, or poll only while the Stream options section is open.

packages/@expo/hub-client/src/types.ts
```diff
@@ -219,8 +234,8 @@ export interface DeviceCapabilities {
   deviceSettings: boolean;
   activity: boolean;
   events: boolean;
-  /** Runtime encoder settings can be read and patched. */
-  streamSettings: boolean;
+  /** Runtime encoder settings that can be read and patched. */
+  streamSettings: DeviceStreamSettingCapabilities;
```
krystofwoldrich, line 238: **[suggestion]** `streamSettings` changes from `boolean` to `false | Partial<Record<...>>`, and `DeviceClient` gains new required members across this stack. Every in-repo consumer is updated, so this is fine inside the monorepo.

RELEASING.md lists `@expo/hub-client` as a published package, but none of the three PRs adds a changeset for it, and its package.json says `"private": true`. If it is published, this is a breaking type change that needs a `@expo/hub-client` changeset. If it is not, RELEASING.md is stale. Worth settling which is true before this stack merges.

.changeset/android-stream-resolution.md
```diff
@@ -0,0 +1,5 @@
+---
+'expo-device-hub': minor
+---
+
+Add runtime Android stream resolution and capture-source controls, use the 60 FPS capture default, expose pipeline diagnostics, and accurately count browser-presented frames while preserving the Hub session across stream restarts.
```
krystofwoldrich: **[nit]** "use the 60 FPS capture default" and "preserving the Hub session across encoder restarts" describe serve-emu behavior (`SCRCPY_DEFAULTS.maxFps` and the router's session handling), not code in this diff. Keep the changeset to what this package changes: the resolution control, the source control, per-key settings capabilities, and presented-frame FPS accounting.

krystofwoldrich (commented):

## Review: Add Android stream resolution and source controls (PR 2/3 in stack: #39 → #40 → #41)

**Verdict:** REQUEST_CHANGES (posted as a comment; the reviewing account is the PR author)
**Findings:** 0 critical · 1 design · 4 suggestion · 1 nit

This PR adds Android clients for serve-emu's `/api/stream-settings` (poll + PATCH) and `/api/stream-mode` (poll + PUT), replaces the boolean `capabilities.streamSettings` with a per-key map so the UI renders only the controls the backend supports, and counts `metadata.presentedFrames` deltas so client FPS includes frames the browser skipped callbacks for. I checked the request and response shapes against serve-emu's `api-contracts.ts` at the pinned commit; they match.

Main concern: the Android settings client re-implements the iOS one almost line for line, and #41 then adds a third copy of the PUT flow.

### Findings

- **[design]** `packages/@expo/hub-client/src/useAndroidDevice.ts:581` — `updateStreamSettings` and the GET poll duplicate `useIosDevice.ts`. Inline.
- **[suggestion]** `packages/@expo/hub-client/src/useAndroidDevice.ts:500` — failed source switch is swallowed silently. Inline.
- **[suggestion]** `packages/@expo/hub-client/src/useAndroidDevice.ts:166` — iOS still counts one frame per callback, so "Client FPS" means different things per platform. Inline.
- **[suggestion]** `packages/@expo/hub-client/src/useAndroidDevice.ts:78` — eight polling loops per Android client. Inline.
- **[suggestion]** `packages/@expo/hub-client/src/types.ts:238` — breaking type change with no `@expo/hub-client` changeset; RELEASING.md and package.json disagree on whether it is published. Inline.
- **[nit]** `.changeset/android-stream-resolution.md:5` — changeset claims serve-emu behavior. Inline.

---

packages/@expo/hub-client/src/types.ts
krystofwoldrich: we dont add changelogs for private packages

.changeset/android-stream-resolution.md
krystofwoldrich: simplify this to one sentence "Add Android stream resolution and source controls"

packages/@expo/hub-client/src/useAndroidDevice.ts
krystofwoldrich: Implemented in d90b2bbc: iOS and Android now use the shared useStreamSettingsResource GET/PATCH state machine with backend-specific parse/toPatch adapters. Shared semantic equality is covered in stream-settings.test.ts. — Codex (GPT-5, high reasoning)

packages/@expo/hub-client/src/useAndroidDevice.ts
krystofwoldrich: Implemented in d90b2bbc: failed source PUTs now preserve serve-emu error detail in streamSourceError, and StreamOptionsSection renders it as an alert. Contract and UI tests cover the detail and HTTP fallback. — Codex (GPT-5, high reasoning)

packages/@expo/hub-client/src/useAndroidDevice.ts
krystofwoldrich: Implemented in d90b2bbc: presentedVideoFrameDelta now lives in a shared module and both Android and iOS WebRTC frame callbacks use metadata deltas. The shared helper has its own unit test. — Codex (GPT-5, high reasoning)

packages/@expo/hub-client/src/useAndroidDevice.ts
krystofwoldrich: Implemented in d90b2bbc: stream settings and source refreshes now share one 3-second stream-options interval. Polling pauses while document.hidden and refreshes on visibility changes. — Codex (GPT-5, high reasoning)

packages/@expo/hub-client/src/types.ts
krystofwoldrich: Confirmed from repository history that @expo/hub-client was intentionally made private in 768542d4, so it does not receive its own changeset. d90b2bbc updates RELEASING.md and .changeset/README.md to document the single public package and how private workspace changes are released. — Codex (GPT-5, high reasoning)

.changeset/android-stream-resolution.md
krystofwoldrich: Updated exactly as requested in 052ff9db: “Add Android stream resolution and source controls.” — Codex (GPT-5, high reasoning)

PR #41, opened 2026-09-02
Add selectable MMAP gRPC streaming

krystofwoldrich:
## Summary

- add CLI flags for selecting the Android gRPC source and explicit PNG or MMAP delivery
- expose PNG/MMAP selection in the Device Hub stream options UI
- integrate nested gRPC/MMAP producer, transport, host-copy, and FPS diagnostics
- share the stream-mode update lifecycle between source and gRPC image-mode changes

This is the top layer of the Device Hub stream stack. It depends on #40 and the gRPC/MMAP APIs from https://github.com/expo/serve-emu/pull/16. The serve-emu pointer was already updated on `main` by #48, and the packaged serve-emu runtime fix landed separately in #51.

## Test plan

- [x] `bun run test`
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run build`
- [ ] verify PNG and MMAP gRPC streaming in the integrated browser

— Codex (GPT-5)


krystofwoldrich:
Local stack-top validation passed: `bun install --frozen-lockfile`, `bun run test`, `bun run typecheck`, `bun run lint`, and `bun run build`. The build includes the packaged CLI smoke test under Node.js, which also passed.

packages/expo-device-hub/package.json
```diff
@@ -38,6 +38,7 @@
     "vendor/serve-sim/LICENSE",
     "vendor/serve-sim/NOTICE",
     "vendor/serve-emu/dist/*.js",
+    "vendor/serve-emu/dist/shared/*.js",
```
krystofwoldrich: **[critical]** This line is a release blocker for `main`, not only for this stack. `main` already vendors serve-emu `2b161b5`, whose `dist/middleware.js` contains `from "./shared/api-contracts.js"`, and `build-plugin-server.ts` keeps that middleware external. With the old non-recursive glob the published tarball has no `dist/shared/`, so `npx expo-device-hub` fails at import time.

Please land this line (with the smoke test) as its own PR against `main` and rebase the stack on it. Consider `vendor/serve-emu/dist/**/*.js` so the next serve-emu subdirectory does not repeat this.

packages/expo-device-hub/package.json
```diff
@@ -82,7 +83,8 @@
     "web": "expo start --web",
     "cli": "bun dist/server/cli.mjs",
     "test": "bun test",
-    "build": "bun run build:web && bun run build:vendor && bun run build:server",
+    "test:package": "bun run scripts/package-smoke.ts",
+    "build": "bun run build:web && bun run build:vendor && bun run build:server && bun run test:package",
```
krystofwoldrich: **[design]** `build` now runs `npm pack` twice, `tar`, and `node` on every `bun run build`, including local iteration and `turbo run typecheck` (which depends on build; I watched it run during typecheck). It also makes `build` depend on `npm` and `tar` being on PATH. The check is valuable, but it is a test. Keep `test:package` as a separate script and run it from `release.yml` and the CI `test` step rather than inside `build`.

packages/@expo/hub-client/src/types.ts
```diff
@@ -216,6 +220,37 @@ export interface DeviceStreamCaptureStats {
   offeredFrames: number | null;
   forwardedFrames: number | null;
   pumpRestarts: number | null;
+  /** Selected emulator screenshot delivery strategy. Null for non-gRPC sources. */
+  grpcImageMode?: DeviceGrpcImageMode | null;
```
krystofwoldrich: **[design]** `DeviceStreamCaptureStats` gains 23 optional flat `grpc*` / `mmap*` fields. serve-emu's own shape is nested (`capture.grpc: GrpcCaptureDiagnostics | null`), and `StreamOptionsSection` has to write `capture.x ?? null` about twenty times to turn `undefined` back into `null`.

A nested `grpc: DeviceGrpcCaptureStats | null` (with the p50/p95 pairs kept as small objects) mirrors the server, makes "no gRPC data" a single check, and keeps the required-field style used by the rest of this interface.

packages/@expo/hub-client/src/useAndroidDevice.ts
```diff
@@ -513,6 +514,54 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
     [streamSourceUrl],
   );
 
+  const setGrpcImageMode = useCallback(
```
krystofwoldrich, line 518: **[design]** `setGrpcImageMode` is `setStreamSource` with a different body and error string: about 45 duplicated lines of request id, controller ref, pending ref, and then/catch/finally. A single `putStreamMode(body)` used by both callbacks removes the duplication and keeps the pending-state bookkeeping in one place.

packages/expo-device-hub/scripts/package-smoke.ts
```diff
@@ -0,0 +1,91 @@
...
+  );
+
+  const packedApiContracts = join(
+    extractedRoot,
+    'package',
+    'vendor',
+    'serve-emu',
+    'dist',
+    'shared',
+    'api-contracts.js',
```
krystofwoldrich: **[suggestion]** The smoke import targets `shared/api-contracts.js` only. The file the Hub loads at runtime is `vendor/serve-emu/dist/middleware.js`. Importing that from the extracted package exercises the whole vendored import graph and catches the next missing file, not only this one. Node also prints `MODULE_TYPELESS_PACKAGE_JSON` here because the extracted package.json has no `type` field; harmless, but noisy in build output.

packages/expo-device-hub/src/server/cli/options.ts
```diff
@@ -31,6 +35,8 @@ Options:
       --mjpeg-quality <quality> MJPEG quality (0.05-1)
       --video-bitrate <bps>  H.264/WebRTC target bitrate (100000-50000000)
       --video-fps <fps>      H.264/WebRTC frame rate (1-120)
+      --stream-source <source> Android capture source: ${ANDROID_STREAM_SOURCES.join(', ')} (default: scrcpy)
+      --grpc-image-mode <mode> gRPC frames: ${GRPC_IMAGE_MODES.join(', ')} (default: png)
```
krystofwoldrich: **[nit]** `--grpc-image-mode` is accepted with the default `scrcpy` source. serve-emu keeps it as the default for a later runtime switch to gRPC, which is reasonable, but this help text reads as if the flag only matters with `--stream-source grpc-screenshot`. A short note such as "(used when the gRPC source is active)" would make that explicit.

krystofwoldrich (commented):

## Review: Add selectable MMAP gRPC streaming (PR 3/3 in stack: #39 → #40 → #41)

**Verdict:** REQUEST_CHANGES (posted as a comment; the reviewing account is the PR author)
**Findings:** 1 critical · 3 design · 1 suggestion · 1 nit

This PR adds `--stream-source` and `--grpc-image-mode` CLI flags, a PNG/MMAP toggle in Stream options, gRPC pipeline diagnostics in the Capture group, and a package smoke test. It also fixes the npm `files` list so `vendor/serve-emu/dist/shared/*.js` ships.

**The `files` fix is the most important change in the whole stack, and it is in the wrong place.** `main` already points serve-emu at `2b161b5` (#48). The built `dist/middleware.js` at that commit imports `./shared/api-contracts.js`, `build-plugin-server.ts` keeps that middleware external, and `main`'s glob `vendor/serve-emu/dist/*.js` is non-recursive. A release cut from `main` today ships a CLI that fails at import time. Sitting at the top of a three-PR feature stack, the fix is blocked behind two feature reviews.

Two statements in the PR body do not match the diff: the serve-emu submodule pointer is unchanged in all three PRs (`main` is already at the PR #16 tip), and there is no `Bun is not defined` change in this diff.

### Findings

- **[critical]** `packages/expo-device-hub/package.json:41` — release-blocking `files` fix buried at the top of the stack. Inline.
- **[design]** `packages/expo-device-hub/package.json:87` — `build` now runs `npm pack` twice, `tar`, and `node` on every build. Inline.
- **[design]** `packages/@expo/hub-client/src/types.ts:224` — 23 flat optional `grpc*`/`mmap*` fields where a nested object mirrors the server and removes ~20 `?? null` in the UI. Inline.
- **[design]** `packages/@expo/hub-client/src/useAndroidDevice.ts:517` — `setGrpcImageMode` duplicates `setStreamSource`. Inline.
- **[suggestion]** `packages/expo-device-hub/scripts/package-smoke.ts:76` — smoke test imports `api-contracts.js` instead of the runtime entry `middleware.js`. Inline.
- **[nit]** `packages/expo-device-hub/src/server/cli/options.ts:39` — help text for `--grpc-image-mode` does not say it applies with the default source too. Inline.

---

packages/expo-device-hub/package.json
krystofwoldrich: Addressed in #51, targeting `main`. The fix uses `vendor/serve-emu/dist/**/*.js` and declares the compatible Busboy runtime dependency. — Codex (GPT-5)

packages/expo-device-hub/package.json
krystofwoldrich: We decided to drop the test:package, this will be solved later. By refactor how the packages are vendored.

packages/expo-device-hub/scripts/package-smoke.ts
krystofwoldrich: We will drop this.

PR #43, opened 2026-09-02
Refactor Android device settings into one table

vonovak:
## Summary

- replace the three parallel `if`-chains in `android-device-settings.ts` with one `Record<AndroidDeviceSettingKey, AndroidDeviceSettingSpec>`
- derive `ANDROID_DEVICE_SETTING_KEYS`, the polled subset, and the per-key version map from that table
- no behavior change

## Context

`androidDeviceSettingPath`, `androidDeviceSettingRequest` and `parseAndroidDeviceSetting` each branched on the same key names, and `useAndroidDevice.ts` repeated the key list three more times. Adding one setting meant editing six places, and missing the version map silently read `undefined` in the poll's version check. A `Record` over the closed union makes a missing key a compile error instead.

serve-sim already proved this shape with its `UI_OPTIONS` registry, which is why the iOS path is option-agnostic while the Android path was not. This brings Android to the same shape.

This layer keeps `main`'s serve-emu submodule pin and adds no route, so unlike the layers above it does not depend on expo/serve-emu#26 and can merge on its own.

## Verification

- `bun run typecheck` — 11 tasks successful
- `bun run test` — 522 passed, 0 failed (167 in `expo-device-hub`)
- `bun run lint` — 0 warnings, 0 errors
- `packages/expo-device-hub` defines no `typecheck` script, so the root gate never reads its source. `turbo run typecheck --dry-run=json` reports `expo-device-hub#typecheck -> <NONEXISTENT>`. Ran `tsc --noEmit` in that package directly: 206 error headers, all 206 in `vendor/serve-emu/` (mostly TS5097 on `.ts` import extensions in the vendored copy), none in `src`. This gap predates the stack.

The assertions in `android-device-settings.test.ts` are untouched by this PR, which is the evidence that behavior is preserved.

## Rebase note

Rebased onto `main` after #55 and #67. One conflict in `useAndroidDevice.ts`: `main` renamed `DeviceSettingWriteTracker` to `KeyedWriteTracker<DeviceSettingKey>` and `deviceSettingScope` to `deviceScope`. Resolved by keeping `main`'s names alongside this PR's `createAndroidDeviceSettingVersions()`.

### DIFF
--- a/packages/@expo/hub-client/src/android-device-settings.ts
+++ b/packages/@expo/hub-client/src/android-device-settings.ts
@@ -30,61 +30,105 @@ export function androidTextSizeForFontScale(value: unknown): AndroidTextSize | n
   ).value;
 }
 
+export type AndroidDeviceSettingPath = '/api/uimode' | '/api/network' | '/api/font-scale';
+
 export interface AndroidDeviceSettingRequest {
-  path: '/api/uimode' | '/api/network' | '/api/font-scale';
+  path: AndroidDeviceSettingPath;
   body: Record<string, unknown>;
 }
 
-export function androidDeviceSettingPath(
+function asRecord(value: unknown): Record<string, unknown> | null {
+  if (!value || typeof value !== 'object' || Array.isArray(value)) return null;
+  return value as Record<string, unknown>;
+}
+
+interface AndroidDeviceSettingSpec {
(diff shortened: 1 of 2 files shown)


krystofwoldrich (approved):
Good catch. This is more maintainable.

PR #44, opened 2026-09-02
Add Android reduce-motion and increase-contrast controls

vonovak:
## Summary

- bump the serve-emu submodule to pick up `/api/reduce-motion` and `/api/high-text-contrast`
- add both keys to the Android device-settings table
- drop the `platform === 'ios'` gate on the accessibility switches so Android shows the switches it reports
- add a minor changeset

## Depends on

**expo/serve-emu#26.** This PR pins serve-emu `55045f6`, the head of that PR. It cannot merge until #26 lands and the pin moves to the merged SHA.

## Context

Both settings are framework-level, so any app observes them, not only React Native. Reduce motion drives the three `global` animation scales that Android's own "Remove animations" toggle moves, and reads back `transition_animation_scale` as the authority, because that is the single key React Native's `AccessibilityInfo.isReduceMotionEnabled` parses. Increase contrast drives `secure high_text_contrast_enabled`, which `isHighTextContrastEnabled` reads.

### Where the routes live

This PR originally added both routes to an in-repo compatibility layer, `serve-emu-device-options.ts`, because the Expo branch of serve-emu had neither `/api/network` nor `/api/font-scale` at the time. That premise is gone. #39 deleted that file once those two landed upstream, and these four follow the same path: they live in serve-emu (expo/serve-emu#26), and the Hub reaches them through the same vendored router as every other route.

So this layer carries **no server code at all** — only the submodule bump plus the client and UI wiring. That also removes the duplication the compatibility layer required: it shipped its own `adb` exec wrapper, timeouts, concurrency limiter and JSON body reader beside the ones serve-emu already had.

Colour filter and invert colours are deliberately excluded. Both accept writes and are readable by apps, but `dumpsys color_display` reports the display features as "Not available" on emulator images, and screenshot comparison confirms nothing changes on screen. A switch that visibly does nothing reads as broken.

The `platform === 'ios'` removal is a deletion rather than a new Android branch because `visible(key)` already hides any key the backend does not report.

## Verification

- `bun run typecheck` — 11 tasks successful
- `bun run test` — 530 passed, 0 failed (169 in `expo-device-hub`)
- `bun run lint` — 0 warnings, 0 errors
- `packages/expo-device-hub` defines no `typecheck` script, so the root gate never reads its source. `turbo run typecheck --dry-run=json` reports `expo-device-hub#typecheck -> <NONEXISTENT>`. Ran `tsc --noEmit` in that package directly: 206 error headers, all 206 in `vendor/serve-emu/` (mostly TS5097 on `.ts` import extensions in the vendored copy), none in `src`. This gap predates the stack.
- Wire contract checked against the built bundle rather than the source: `/api/reduce-motion` and `/api/high-text-contrast`, and their `reduceMotion` and `highTextContrast` response keys, all appear in `vendor/serve-emu/dist/middleware.js`.
- End to end against a booted API 36 `google_apis_playstore` emulator, driven through the browser rather than the API: clicked both switches, confirmed via `adb` that all three animation scales became `0` and `high_text_contrast_enabled` became `1`, confirmed both switches read back as checked, then confirmed toggling off restored the defaults. That run predates the move to serve-emu. The adb commands and the response envelopes are unchanged, so it now stands as evidence for serve-emu's implementation of the same contract.

### DIFF
--- a/.changeset/android-accessibility-options.md
+++ b/.changeset/android-accessibility-options.md
@@ -1,0 +1,5 @@
+---
+'expo-device-hub': minor
+---
+
+Add Android reduce-motion and increase-contrast controls to the shared device options UI.
--- a/packages/@expo/hub-client/src/__tests__/android-device-settings.test.ts
+++ b/packages/@expo/hub-client/src/__tests__/android-device-settings.test.ts
@@ -1,9 +1,12 @@
 import { describe, expect, test } from 'bun:test';
 
 import {
+  ANDROID_DEVICE_SETTING_KEYS,
+  ANDROID_POLLED_DEVICE_SETTING_KEYS,
   androidDeviceSettingRequest,
   androidFontScaleForTextSize,
   androidTextSizeForFontScale,
+  createAndroidDeviceSettingVersions,
   parseAndroidDeviceSetting,
 } from '../android-device-settings';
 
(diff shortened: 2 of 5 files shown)


.changeset/android-accessibility-options.md
```diff
@@ -0,0 +1,7 @@
+---
+'@expo/hub-client': minor
+'@expo/hub-components': minor
```
krystofwoldrich: Only expo-device-hub is public, the other package can't have changelog. The release CI would fail on it.
 
````suggestion
````

File: packages/expo-device-hub/src/server/serve-emu-device-options.ts [hunk 2]
```diff
@@ -274,5 +368,62 @@ export async function handleServeEmuDeviceOptionRequest(
     return new Response('method not allowed', { status: 405 });
   }
 
+  if (pathname === '/api/reduce-motion') {
```
krystofwoldrich: Looks good in context of the existing code. 

Ideally we move these to https://github.com/expo/serve-emu/blob/expo/packages/serve-emu/src/api/routes/devices.ts

krystofwoldrich (approved):

PR #50, opened 2026-09-03
fix(android): keep the decoder across encoder restarts

vonovak:
## Why

Two defensive changes for the Android stream. Neither changes measured behavior against current `main`; see Verification.

- serve-emu sends a `video-session` message after an encoder restart. The Hub client closed its decoder and requested a keyframe on every such message, which can loop when the server restarts the encoder in response to that request. This branch keeps the decoder unless the reported size changed and requests a keyframe only on the first session message of a connection.
- The keyframe interval passed to the vendored serve-emu is now explicit at 10 s. The vendored revision already defaults to 10 s, so this only protects against a future revision that changes the default.

## Scope

- `useAndroidDeviceClient` in `packages/@expo/hub-client/src/useAndroidDevice.ts`: `video-session` handling as above. WebSocket transport only; the WebRTC path does not handle that message.
- `standaloneServeEmuOptions` and the fallback options in `packages/expo-device-hub/src/server/serve-emu-options.ts`: `keyFrameInterval: 10`.
- One changeset for `expo-device-hub`.

## Tradeoffs

- A client that misses frames recovers through the rate-limited `reset-video` request rather than a scheduled keyframe. With the 10 s interval that is already the case on `main`, so this branch does not change recovery time.
- The client now trusts the server to send a keyframe after a restart instead of asking for one on each `video-session`. The vendored serve-emu starts every restart with a keyframe and marks the client as waiting for one. A server that restarted without a keyframe would leave the client waiting until it saw a delta frame and asked.
- A size change still tears the decoder down, so rotation keeps its reconfigure path.

## Blast Radius

Android streams through the Hub over the WebSocket transport. WebRTC and iOS are untouched.

## Verification

Paired runs of the Hub CLI built from `origin/main` at `bc1d36d` and from this branch, same harness, same Pixel 8a AVD on an M1 Max with the host GPU, one emulator on the host, two rounds each. A Playwright script drives headless Chrome against the real dashboard, a probe APK animates a box and toggles the screen on tap, and hooks in the page count decoder teardowns and `reset-video` requests over a 15 s motion window.

| Source | Tree | Decoder teardowns | Keyframe requests | Stream fps (median, p10) | Tap-to-glass (p50, p95) |
|---|---|---|---|---|---|
| gRPC mmap (default) | `main` | 0, 0 | 0, 0 | 23, 22 | 116 ms, 133 ms |
| gRPC mmap (default) | branch | 0, 0 | 0, 0 | 23, 21 | 114 ms, 126 ms |
| scrcpy | `main` | 0, 0 | 0, 0 | 29, 25 | 69 ms, 85 ms |
| scrcpy | branch | 0, 0 | 0, 0 | 30, 29 | 71 ms, 84 ms |

Every received frame was painted in every run. The differences are within run-to-run variance. The looping behavior this branch guards against did not occur on `main` in these runs: three `client opened` resets per run, all from connection setup, none from the client's handler. I observed the loop earlier on a different build of this branch, 11 teardowns and 11 requests in 15 s, but I could not reproduce it against `main` and have not identified what that build did differently.

Tests: `bun test` in `packages/@expo/hub-client` (135 pass) and `packages/expo-device-hub` (154 pass). `bun run typecheck` at the root passes.

### DIFF
--- a/.changeset/android-keyframe-interval.md
+++ b/.changeset/android-keyframe-interval.md
@@ -1,0 +1,5 @@
+---
+'expo-device-hub': patch
+---
+
+Ask the Android encoder for a keyframe every 10 seconds instead of every second. Late joiners still get one on demand, and the stream no longer carries a per-second burst.
--- a/packages/@expo/hub-client/src/useAndroidDevice.ts
+++ b/packages/@expo/hub-client/src/useAndroidDevice.ts
@@ -841,6 +841,7 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
     let decoder: VideoDecoder | null = null;
     let sawKeyframe = false;
     let droppingUntilKeyframe = false;
+    let sessionSize: ScreenSize | null = null;
     let lastKeyframeRequestAt = 0;
     let frameIdx = 0;
     let fpsCount = 0;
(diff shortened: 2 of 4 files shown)


vonovak:
closing as not needed, https://github.com/expo/serve-emu/commit/9f859eb558c8f6a15395c3315e6768d6ab800253 already sets this

PR #54, opened 2026-09-03
Keep the Android frame live while the stream source switches

krystofwoldrich:
Switching the Android capture source (scrcpy ↔ gRPC), the gRPC input source, or the gRPC image mode (PNG ↔ MMAP) in the Hub blanked the device frame and let the sidebar run ahead of the stream. serve-emu's own UI keeps the last frame and simply continues; this PR makes the Hub behave the same way.

Rebased onto #55 and #56, so the fix covers the new Input source select as well.

## Before

serve-emu stages the new capture session, publishes it, closes every viewer socket of the old session (code 1001, `server stopping`), and only then answers `PUT /api/stream-mode`. The Hub client treated that close as a disconnect:

- The device frame turned black under a red "Disconnected — retrying…" overlay (WebSocket) or "WebRTC input disconnected. Retrying..." (WebRTC), and the status pill said "Error".
- The sidebar controls flipped to the new source as soon as the PUT answered, 0.5–2 s before the replacement stream painted.
- The reconnect waited out the 500 ms backoff before dialing again.

Timed with an in-page probe in the integrated browser (WebSocket transport, MMAP → PNG): socket closed at +1.79 s, sidebar flipped at +1.81 s, frame live again at +4.02 s, so about 2.2 s of black frame. scrcpy ↔ gRPC showed the same pattern with about 1.3 s of black frame. The WebRTC transport kept its video flowing but lost the control socket, and the Hub still covered the frame with the error overlay for about 2 s.

## After

Mid-switch (t = 2.4 s): before on the left, after on the right

The frame never blanks and never shows an overlay. The controls stay disabled with a "Switching stream source…" note, and they adopt the new value in the same frame the replacement stream becomes live.

Re-verified after the rebase against the redesigned sidebar, on `expo-emu-host-0`:

| Switch | Controls disabled | Pill "Reconnecting" | Frame live + controls flip |
|---|---|---|---|
| gRPC → scrcpy | +0.26 s | +1.46 s | +1.81 s |
| scrcpy → gRPC | +0.26 s | +1.11 s | +1.27 s |
| Input scrcpy → gRPC | +0.26 s | +1.00 s | +1.24 s |
| MMAP → PNG | +0.26 s | +2.38 s | +3.93 s |

Recordings, captured before the rebase on the pre-redesign sidebar (Hub CLI + `expo-emu-host-0`, WebSocket transport, headless Chrome screencast):

- [before-websocket.mp4](https://github.com/expo/expo-device-hub/blob/1b99f6070d211e9d4471c1652dec3dae3cfe9b03/assets/pr/android-stream-switch/before-websocket.mp4)
- [after-websocket.mp4](https://github.com/expo/expo-device-hub/blob/1b99f6070d211e9d4471c1652dec3dae3cfe9b03/assets/pr/android-stream-switch/after-websocket.mp4)

## Changes

`@expo/hub-client`

- New `reconnecting` connection status. A stream that was live and loses its transport keeps its last frame (canvas pixels, or the ended WebRTC track) and reports "Reconnecting"; it becomes `error` only after a 5 s grace period (`stream-reconnect.ts`).
- A deliberate server close (1000/1001/1012) of a healthy socket retries after 100 ms instead of the 500 ms backoff, for both the video socket and the WebRTC control socket.
- The WebRTC `<video>` keeps its MediaStream while the peer renegotiates; it is detached only when the surface leaves WebRTC or moves to another device.
- `streamSourcePending` covers the whole switch: the request, the old stream dropping, and the first frame of the replacement (`stream-switch.ts`, a pure reducer with safety timeouts). The server's answer is held back until that frame renders, so the Source / Input / gRPC frames controls and the device frame change at the same moment. Polling of `/api/stream-mode` pauses while a switch is pending. Every control routes through `putStreamMode`, so the new input-source select is covered too.
- `DeviceScreen` no longer paints a black background or the status overlay while `reconnecting`.

`@expo/hub-components`

- `DeviceTitle` renders `reconnecting` as a warning-coloured "Reconnecting" pill.
- `StreamOptionsSection` shows a "Switching stream source…" note under the capture controls while a switch is pending; the previous values stay selected until it completes.

The iOS client never emits `reconnecting`, so its behaviour is unchanged.

## Test plan

- [x] `bun install --frozen-lockfile`
- [x] `npx turbo run build --filter=expo-device-hub`
- [x] `bun run test`
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] Integrated browser against the standalone Hub CLI with `expo-emu-host-0`: timed all four capture switches after the rebase (numbers above) plus scrcpy ↔ gRPC and PNG ↔ MMAP on the WebRTC transport before it; browser console stayed free of errors


krystofwoldrich-agent:

## Review: Keep the Android frame live while the stream source switches

**Verdict (at 1b99f60):** REQUEST_CHANGES
**Findings:** 0 critical · 2 design · 0 suggestion · 2 nit

The PR adds a `reconnecting` status, a shared reconnect policy, and a pure switch-tracking reducer. The Android hook holds the server's answer back until the replacement stream paints. The core design is sound. Server side, confirmed in the serve-emu submodule: a source switch closes viewer sockets with code 1012, and a server stop uses code 1001. serve-emu runs in-process in the Hub, so those codes reach the browser unchanged. Typecheck, lint, and all tests pass locally.

### Findings

#### [design] A switch from a dead stream keeps the controls pending for 3 s after the new frame is live
`packages/@expo/hub-client/src/stream-switch.ts:88`
The reducer does not know whether the stream was live when the request started. If the user switches while the status is `connecting` or `error`, no `stream-interrupted` event ever fires. The `request-success` event then lands in `awaiting-interruption`. That phase ignores `stream-live` on purpose, so the replacement frame paints but the sidebar keeps the old value and stays disabled until the 3 s timeout in `useAndroidDevice.ts`. This is the same frame-versus-sidebar desync the PR fixes, in the other direction. Switching away from a broken source is a common reason to use these controls. Fix: carry `live` on `request-start` and, when the stream was not live, move to `awaiting-frame` directly or end the switch at once.

#### [design] PR media is committed to the source repository
`assets/pr/android-stream-switch/before-websocket.mp4`
The PR adds two videos and a screenshot, 716 KB total, under `assets/pr/`. On `main` the `assets/` folder holds only the two README banners. These files exist only to illustrate the PR body and will stay in the history of every clone. GitHub can host them when they are dragged into the PR description.

#### [nit] Two consecutive doc blocks on `ConnectionStatus`
`packages/@expo/hub-client/src/types.ts:35`
The existing block lists the four states. The new block for `reconnecting` follows it as a separate comment, so the first block is orphaned. Add `reconnecting` as a line in the existing list.

#### [nit] Header says the policy is shared, but only Android uses it
`packages/@expo/hub-client/src/stream-reconnect.ts:2`
The file says "Reconnect policy shared by the device clients". The iOS client is unchanged and keeps its own inline backoff.

Checked and correct: the grace timer clears on the first painted frame, the WebRTC detach effect only fires on transport or device change, `refreshStreamSource` polling pauses while a switch is pending, and `resetStreamSwitch` runs on device change so stale answers cannot commit.

---

**Follow-up:** all four findings are addressed in 9062839. The PR media was removed from the branch, so the images and videos in the description need to be re-uploaded as attachments.

PR #56, opened 2026-09-03
Add input source selector for gRPC streaming

krystofwoldrich:
## Summary

This fixes keyboard input not working for emulators which are started with hw keyboard setting disabled. `scrcpy` input works regardless the emulator settings.

- show an Input selector when the Android stream source is gRPC
- default gRPC video to scrcpy input, with gRPC input still selectable
- keep input source and gRPC frame mode consistent across stream restarts
- pin the serve-emu transport implementation from https://github.com/expo/serve-emu/pull/20

## Testing

- bun run typecheck
- bun run test
- bun run lint
- serve-emu: bun run check
- integrated Codex browser against Android emulator: switched scrcpy → gRPC → scrcpy input, then triggered Home; stream remained live

## Screenshot

Expo Device Hub showing gRPC streaming with scrcpy input selected

— Codex (GPT-5.6)

### DIFF
--- a/packages/@expo/hub-client/src/useAndroidDevice.ts
+++ b/packages/@expo/hub-client/src/useAndroidDevice.ts
@@ -56,6 +56,7 @@ import {
   type DeviceConnectionOptions,
   type DeviceEvent,
   type DeviceGrpcImageMode,
+  type DeviceInputSource,
   type DeviceLog,
   type DeviceSettingKey,
   type DeviceSettings,
@@ -448,7 +449,11 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
   });
 
   const putStreamMode = useCallback(
-    (body: { mode: DeviceStreamSource; grpcImageMode?: DeviceGrpcImageMode }) => {
+    (body: {
+      mode: DeviceStreamSource;
+      grpcImageMode?: DeviceGrpcImageMode;
+      inputSource?: DeviceInputSource;
+    }) => {
       if (!streamSourceUrl || streamSourcePendingRef.current) return;
       const request = ++streamSourceRequestRef.current;
       const controller = new AbortController();
@@ -510,7 +515,10 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
       ) {
         return;
       }
-      putStreamMode({ mode: source });
+      putStreamMode({
+        mode: source,
+        ...(source === 'grpc-screenshot' ? { inputSource: 'scrcpy' as const } : {}),
+      });
     },
     [putStreamMode],
   );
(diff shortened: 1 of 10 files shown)


packages/@expo/hub-client/src/useAndroidDevice.ts
```diff
@@ -510,7 +515,10 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
       ) {
         return;
       }
-      putStreamMode({ mode: source });
+      putStreamMode({
+        mode: source,
+        ...(source === 'grpc-screenshot' ? { inputSource: 'scrcpy' as const } : {}),
```
krystofwoldrich-agent: **[suggestion]** `setStreamSource` always sends `inputSource: 'scrcpy'` when switching to gRPC video. serve-emu already defaults a fresh device to scrcpy input, and when the field is omitted it restores the last per-device choice (`inputSourceOverrides` in the router, `inputSources` in the single-device server). So this line is redundant for the default case, and it discards a user's earlier gRPC-input selection when they toggle scrcpy → gRPC video. `grpcImageMode` is not sent here and *is* restored from the server's memory, so the two sibling settings now behave differently on the same switch. If the reset is intentional, a comment would help; otherwise omit the field so both settings follow the server's remembered value.

krystofwoldrich-agent (changes requested):

## Review: Add input source selector for gRPC streaming

**Verdict:** REQUEST_CHANGES
**Findings:** 0 critical · 1 design · 1 suggestion · 0 nit

This PR extends the hub-client stream-source contract with `inputSource` and `availableInputSources`, adds `setGrpcInputSource` to `DeviceClient`, renders an "Input" segmented control in `StreamOptionsSection` when gRPC video is active and more than one input is available, and pins the `serve-emu` submodule to the head of expo/serve-emu#20.

The code changes are sound and follow the existing `grpcImageMode` pattern throughout. Typecheck, tests (154 pass in `expo-device-hub`), and lint pass locally with the submodule at `0eae30a`. The only blocker is the submodule pin.

### Findings

#### [design] Submodule pinned to an unmerged feature-branch commit
`packages/serve-emu` → `0eae30ad508780e003d08dd6d439073bfea4446d`
This commit is only reachable from `codex/grpc-scrcpy-input` on expo/serve-emu, not from the `expo` branch that `.gitmodules` tracks. `expo` already has two commits past the merge-base (`1ec8462`), so merging expo/serve-emu#20 will produce a different SHA. If that PR is squash-merged and the branch is deleted, the pinned commit becomes unreachable and `git submodule update --init` fails for fresh clones and CI. `bun run build:vendor` also builds the published hub from this checkout. Merge expo/serve-emu#20 first, then re-pin to the resulting SHA on `expo` before merging this PR.

#### [suggestion] `setStreamSource` overrides the server's remembered input choice
`packages/@expo/hub-client/src/useAndroidDevice.ts:520` (inline)

packages/@expo/hub-client/src/useAndroidDevice.ts
krystofwoldrich: Fixed in bc5eefe0. `setStreamSource` now sends only the requested capture mode, so serve-emu restores both the remembered gRPC image mode and the remembered input source when the user switches back to gRPC video. Fresh devices still default to scrcpy input on the server.

— Codex (GPT-5.6)

krystofwoldrich:
Updated the submodule pin to `5483995`, the squash-merge commit for expo/serve-emu#20 on the tracked `expo` branch. Also addressed the input-source suggestion in `bc5eefe0`. Parent typecheck, tests, and lint all pass.

— Codex (GPT-5.6)

PR #62, opened 2026-09-07
fix(expo-device-hub): wait for the emulator to leave adb before shutdown reports success

vonovak:
## Why

`adb emu kill` returns in under 20 ms while the emulator is still exiting, and adb keeps serving the dying serial for around ten seconds. A boot started in that window fails with "Failed to allocate an emulator port", because `freeEmulatorPort` lists devices and `getprop` against the exiting serial errors. Shutdown followed by a boot from the picker hits this today on `main`.

## Scope

- `@expo/hub-android-utils`: new `waitForAdbOffline(serial, timeoutMs)`, exported next to `waitForAdbOnline`. It polls `adb devices -l` until the serial is gone or the timeout elapses. Each listing runs with the remaining budget and the abort signal, so a stalled adb cannot hold the wait open, and a listing that arrives after the deadline or an abort is discarded rather than accepted. `runAdbDevices` takes `timeoutMs` and `signal` and passes them to `execFile`, so the child is killed too.
- `expo-device-hub` server: `shutdownHubDevice` and `removeHubDevice` call `shutdownAndroidHubDevice`, which runs `adb emu kill` and then waits up to 30 seconds for adb to release the serial. A serial that stays listed makes the action fail with a timeout error instead of reporting success.

The wait reads only the raw listing. `listDevices` runs a `getprop` per device and fails the whole listing when any one device is locked or on its way out, which is exactly the window this waits through.

## Blast Radius

Android shutdown and remove now block for up to 30 seconds while the emulator exits, and fail instead of reporting success if it does not. iOS paths are untouched. The boot command is unchanged.

## Verification

- `bun test` in `hub-android-utils` (118, with cases for a listing that never resolves, an abort during a listing, an empty listing after the deadline, and the child timeout) and `expo-device-hub` (161): all pass. `bun run typecheck` and `bun run lint` at the root pass.
- Runtime, on the Hub dev server: `POST /api/devices/shutdown` for a running emulator returned `ok: true` after 4.6 seconds, and `adb devices` was already empty when it returned. A `POST /api/devices/boot` for the same AVD fired right after came online on the same serial. On `main` that second call fails with "Failed to allocate an emulator port".

The camera passthrough PR stacks on this one.

### DIFF
--- a/packages/@expo/hub-android-utils/src/wait-for-adb-offline.ts
+++ b/packages/@expo/hub-android-utils/src/wait-for-adb-offline.ts
@@ -1,0 +1,59 @@
+import { homedir } from "node:os";
+import { runAdbDevices } from "./adb";
+import { type AndroidUtilsError, type AndroidUtilsResult, reportError, result } from "./errors";
+import { parseAdbDevices } from "./parse-adb-devices";
+import { resolveAdbPath } from "./sdk-paths";
+
+export const SHUTDOWN_POLL_INTERVAL_MS = 1500;
+
+async function listAdbSerials(): Promise<AndroidUtilsResult<string[]>> {
+  const adb = resolveAdbPath(process.env, homedir());
+  const listed = await runAdbDevices(adb);
+  if (listed.error) return result([], listed.error);
+  return result(listed.value ? parseAdbDevices(listed.value).map((device) => device.serial) : []);
+}
+
+export interface WaitForAdbOfflineOptions {
+  listSerialsFn?: () => Promise<AndroidUtilsResult<string[]>>;
+  pollIntervalMs?: number;
+  signal?: AbortSignal;
+}
+
+/**
+ * Poll `adb devices -l` until `serial` is gone from it, or time out.
+ *
+ * Its result `value` becomes `true` on the first successful listing that no
+ * longer carries `serial`, or `false` once the timeout elapses or the signal
+ * aborts. A failed listing does not end the wait; the last failure is returned
+ * in `error` alongside `false`.
+ */
+export async function waitForAdbOffline(
+  serial: string,
+  timeoutMs: number,
+  {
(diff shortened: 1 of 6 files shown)


packages/@expo/hub-android-utils/src/wait-for-adb-offline.ts
```diff
@@ -0,0 +1,59 @@
...
+    pollIntervalMs = SHUTDOWN_POLL_INTERVAL_MS,
+    signal,
+  }: WaitForAdbOfflineOptions = {},
+): Promise<AndroidUtilsResult<boolean>> {
+  const deadline = Date.now() + timeoutMs;
+  let lastError: AndroidUtilsError | null = null;
+  while (Date.now() < deadline) {
+    if (signal?.aborted) return result(false, lastError);
+    try {
+      const listed = await listSerialsFn();
```
krystofwoldrich: Bound the in-flight adb listing by the shutdown deadline

The deadline and AbortSignal are only checked before this await. Since runAdbDevices() starts the adb child process without a timeout or signal, a stalled `adb devices -l` leaves shutdown pending indefinitely, even after the 30-second budget expires or the signal aborts. A listing that completes after the deadline can also incorrectly return success because the result is accepted without rechecking the deadline or cancellation.

Please bound the child process by the remaining shutdown budget, propagate cancellation to it, and recheck the deadline/signal after the await before accepting success. Add coverage for a never-resolving listing, cancellation during an in-flight listing, and an empty listing arriving after the deadline. This is independent of the emulator itself refusing to exit.

— Codex (GPT-6)

krystofwoldrich: Just want my agent noticed, I believe we have this issue with most of the ADB calls at the moment. So also fine, if we solve this in a later PR.

krystofwoldrich (approved):

PR #22, opened 2026-08-25
fix(hub-client): honor serve-sim WebRTC settings

krystofwoldrich-agent:
## Summary

- preserve serve-sim advertised WebRTC codec and validated ICE/TURN configuration in the Hub iOS client
- keep codec fallback anchored to the configured or viewer-selected codec
- expose H.264, VP9, and VP8 controls in the Device Hub sidebar and reset the FPS sample window on stream changes
- add patch changesets for @expo/hub-client and expo-device-hub

## Testing

- bun test in packages/@expo/hub-client (55 passed)
- bun test in packages/expo-device-hub (68 passed)
- turbo typecheck for @expo/hub-client and @expo/hub-components
- bun run build:web in packages/expo-device-hub

PR #63, opened 2026-09-07
feat(expo-device-hub): add a Camera section for Android emulators

vonovak:
## Why

serve-emu can now feed an Android emulator's camera from a PNG on the host (expo/serve-emu#18, with the middleware routes in expo/serve-emu#22). The Hub boots emulators itself, so nothing in serve-emu can attach the feeds for it. This PR does the wiring on the Hub side and gives the inspector a Camera section, so a person or an agent can hand the app under test a known picture without leaving the dashboard.

The feed path is fixed at emulator launch. That is why every emulator the Hub boots gets the feeds attached, with no toggle. An opt-in would leave the panel dead until a restart, and serve-emu's test card is a fair default for an app under test.

## Scope

- `packages/serve-emu` is pinned to `dbd0174`, the head of expo/serve-emu#22, which stacks on #18. Re-pin once both land.
- `@expo/hub-android-utils`: `BootDeviceOptions.extraArgs` appends emulator flags after `-port`. `emulatorSerial` is exported.
- `expo-device-hub` server: `bootAndroidHubDevice` seeds both feeds, boots with `cameraLaunchArgs(serial)`, and calls `router.setCameraWired(serial, true)` right after the spawn. The claim clears when the process exits, through `booted.exited`, and on shutdown and remove. A seed failure boots the emulator without the camera flags and reports the failure in `errors`. `device-actions.ts` depends on one `EmulatorCameraFeeds` collaborator, and `serve-emu.ts` supplies it from the vendored middleware.
- `@expo/hub-client`: `useAndroidCamera` polls `GET /api/camera` and the Android client exposes `camera`, `cameraPending`, `cameraError`, `setCameraImage`, and `clearCameraImage` on `DeviceClient`. `capabilities.camera` is true once the backend reports a supported emulator. `parseAndroidCameraStatus` is the one trust boundary and `applyCameraRead` folds every read into the held state, so a failed read keeps the last good status. `KeyedWriteTracker` in `keyed-write-tracker.ts` replaces `DeviceSettingWriteTracker` so device settings and camera facings share it. `apiUrl` and `deviceApiUrl` move to `android-api-url.ts`.
- `@expo/hub-components`: `CameraSection` renders one 4:3 preview per facing with a PNG picker, a drop target, and Reset. `LogSidebar` places it between Stream options and Events. `SectionNote` moves out of `StreamOptionsSection`.

Out of scope: iOS. serve-sim's camera is CLI only and has no mutating middleware route yet.

## Tradeoffs

- Camera writes are not optimistic. The backend's response is the authoritative status. A rejected upload must not show a picture the emulator does not hold, so the UI waits for the answer.
- The wiring claim lives in the router process. A Hub restart loses it, so a wired emulator that outlives the Hub shows as unwired until it is rebooted. Fixing that means deriving the claim in serve-emu's `readCameraStatus`, which is an upstream follow-up on expo/serve-emu#22.
- An emulator started outside the Hub has no wiring claim. The section still shows, with the controls disabled and a note to boot it from the Hub, instead of accepting uploads the emulator would never display.

## Blast Radius

Every Android emulator the Hub boots now runs with `-camera-back imagefile:… -camera-front imagefile:…`. Its camera shows serve-emu's checkerboard test card instead of the emulator's virtual scene until an image is posted. Emulators booted elsewhere, physical devices, and iOS simulators are untouched. `DeviceClient` gains required members, so any consumer that builds one by hand (the website) must add them.

## Verification

- `bun test` in `hub-android-utils` (106), `hub-client` (163), `hub-components` (23), and `expo-device-hub` (171): all pass. `bun run typecheck` and `bun run lint` at the root: 11 and 4 tasks pass. `bun run build:server` bundles.
- Ran the Hub dev server from this worktree with the vendored serve-emu at the pinned commit. Booted the `serve_emu_cam` AVD through `POST /api/devices/boot`. The emulator process carries both camera flags and `GET /vendor/serve-emu/api/camera?device=emulator-5556` reports `wiredAtLaunch: true`.
- A script against the Hub mount: status reads, a posted 1280x960 PNG comes back with a matching digest and identical bytes, a JPEG body is refused with 400 and the other feed stays untouched.
- Opened the camera app on that emulator over adb and captured the screen. The preview shows the posted fixture.
- In the dashboard, the Camera section for that emulator showed the posted back image and the front test card with the controls enabled. A drop of a PNG onto the front box updated it. A drop of a JPEG showed the backend's PNG-only message as an alert. A click on Reset restored the test card.
- For the emulator booted outside the Hub, the section showed the unwired note with every control disabled.
- After `POST /api/devices/shutdown`, the status for that serial reports `wiredAtLaunch: false`.
- Exit watcher: booted an emulator through the Hub (claim true), killed it with `adb emu kill` outside the Hub, then started a flagless emulator on the same serial outside the Hub. The status for that serial reports `wiredAtLaunch: false`.
- A three-model adversarial review ran over the diff. Its seven act-on findings are resolved in these commits.

Stacks on #62, which makes shutdown wait for adb to release the serial. Depends on expo/serve-emu#18 and expo/serve-emu#22.

### DIFF
--- a/packages/@expo/hub-android-utils/src/__tests__/emulator.test.ts
+++ b/packages/@expo/hub-android-utils/src/__tests__/emulator.test.ts
@@ -34,6 +34,21 @@ describe("buildEmulatorArgs", () => {
     const args = buildEmulatorArgs({ name: "x", port: 5556 });
     expect(args[args.indexOf("-port") + 1]).toBe("5556");
   });
+
+  test("appends extraArgs verbatim after the port", () => {
+    const args = buildEmulatorArgs({
+      name: "x",
+      port: 5554,
+      extraArgs: ["-camera-back", "imagefile:/tmp/a.png"],
+    });
+    expect(args.slice(-4)).toEqual(["-port", "5554", "-camera-back", "imagefile:/tmp/a.png"]);
+  });
+
+  test("adds nothing when extraArgs is empty or omitted", () => {
+    const plain = buildEmulatorArgs({ name: "x", port: 5554 });
+    expect(buildEmulatorArgs({ name: "x", port: 5554, extraArgs: [] })).toEqual(plain);
+    expect(plain.at(-1)).toBe("5554");
+  });
 });
 
 describe("formatEmulatorCommand", () => {
@@ -41,6 +56,15 @@ describe("formatEmulatorCommand", () => {
     const command = formatEmulatorCommand("/sdk/emulator/emulator", { name: "x", port: 5556 });
     expect(command.startsWith("/sdk/emulator/emulator ")).toBe(true);
     expect(command).toContain("-port 5556");
+  });
+
+  test("includes the extra args", () => {
+    const command = formatEmulatorCommand("/sdk/emulator/emulator", {
+      name: "x",
+      port: 5554,
+      extraArgs: ["-camera-back", "imagefile:/tmp/a.png"],
+    });
+    expect(command.endsWith("-port 5554 -camera-back imagefile:/tmp/a.png")).toBe(true);
   });
 
   test("quotes parts containing whitespace", () => {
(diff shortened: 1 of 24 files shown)


vonovak:
Superseded by a re-stacked PR on the same branch with the hub-client PR as its base. GitHub does not allow a base change on a stacked PR, so this one is closed and reopened as a new PR.

PR #67, opened 2026-09-07
feat(expo-device-hub): attach camera feeds to every emulator the Hub boots

vonovak:
## Why

The Hub boots emulators itself, so nothing in serve-emu can attach the camera feeds for it. This PR pins serve-emu to the camera middleware, attaches the feeds on every Android boot, and adds the changeset. With it the Camera section from the previous PR comes alive.

The feed path is fixed at emulator launch. That is why every emulator the Hub boots gets the feeds attached, with no toggle. An opt-in would leave the panel dead until a restart, and serve-emu's test card is a fair default for an app under test.

## Scope

- `packages/serve-emu` is pinned to `625469a`, the head of expo/serve-emu#22, which stacks on #18. Re-pin once both land.
- `bootAndroidHubDevice` seeds both feeds and boots with `cameraLaunchArgs(serial)`. A seed failure boots the emulator without the flags and reports the failure in `errors`. `device-actions.ts` depends on one `EmulatorCameraFeeds` collaborator with `launchArgs` and `seedPlaceholders`, and `serve-emu.ts` supplies it from the vendored middleware.
- serve-emu reads `wiredAtLaunch` from the running emulator's `hardware-qemu.ini`, so the Hub declares nothing about wiring. A Hub restart, an emulator killed outside the Hub, or a recycled serial cannot produce a stale answer.
- Changeset for `expo-device-hub`, the one published package; the private `@expo/hub-*` packages ship inside it.

## Tradeoffs

Always-on also attaches a camera to AVDs whose `config.ini` declares no camera hardware. serve-emu documents that `-camera-front imagefile:` yields a front camera even with `hw.camera.front=none`. A check on the AVD config before adding the flags would restore that, if Wear or TV profiles matter.

## Blast Radius

Every Android emulator the Hub boots now runs with `-camera-back imagefile:… -camera-front imagefile:…`. Its camera shows serve-emu's checkerboard test card instead of the emulator's virtual scene until an image is posted. Emulators booted elsewhere, physical devices, and iOS simulators are untouched.

## Verification

- `bun test` in `expo-device-hub`: 167 pass. `boot-hub-device.test.ts` pins the boot order, the launch args for the allocated serial, and the seed-failure fallback. `bun run typecheck` and `bun run lint` at the root pass. `bun run build:server` bundles.
- On the Hub dev server: an AVD booted through `POST /api/devices/boot` carries both camera flags and `GET /vendor/serve-emu/api/camera?device=…` reports `wiredAtLaunch: true`. A posted 1280x960 PNG comes back with a matching digest, a JPEG body is refused with 400, and the emulator's camera app shows the posted image on screen.
- A Hub server restarted under the running emulator still reports `wiredAtLaunch: true`. A dead serial answers 503.

Last PR of the emulator camera stack. Depends on expo/serve-emu#18 and expo/serve-emu#22.

### DIFF
--- a/.changeset/emulator-camera-feeds.md
+++ b/.changeset/emulator-camera-feeds.md
@@ -1,0 +1,5 @@
+---
+'expo-device-hub': minor
+---
+
+Add camera image passthrough for Android emulators.
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ with it, and manage which devices are running from one place.
 - Interact directly — tap, swipe, scroll, and type into the device.
 - Boot, shut down, and add devices without opening Xcode or Android Studio.
 - Follows your system light/dark theme, and can flip the device's appearance too.
+- Feed an Android emulator's camera a PNG from the inspector's Camera section.
 
 > iOS simulators require macOS with Xcode. Android emulators require the Android SDK
 > (`emulator`, `adb`).
--- a/packages/expo-device-hub/README.md
+++ b/packages/expo-device-hub/README.md
@@ -21,6 +21,7 @@ with it, and manage which devices are running from one place.
 - Interact directly — tap, swipe, scroll, and type into the device.
 - Boot, shut down, and add devices without opening Xcode or Android Studio.
 - Follows your system light/dark theme, and can flip the device's appearance too.
+- Feed an Android emulator's camera a PNG from the inspector's Camera section.
 
 > iOS simulators require macOS with Xcode. Android emulators require the Android SDK
 > (`emulator`, `adb`).
--- a/packages/expo-device-hub/src/server/__tests__/boot-hub-device.test.ts
+++ b/packages/expo-device-hub/src/server/__tests__/boot-hub-device.test.ts
@@ -1,0 +1,172 @@
+import { beforeEach, describe, expect, mock, test } from 'bun:test';
+
+import type {
+  AndroidUtilsResult,
+  BootDeviceOptions,
+  BootedDevice,
+  EmulatorExit,
+} from '@expo/hub-android-utils';
+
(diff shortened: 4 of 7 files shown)


.changeset/emulator-camera-feeds.md
```diff
@@ -0,0 +1,8 @@
+---
+'expo-device-hub': minor
+'@expo/hub-client': minor
+'@expo/hub-components': minor
+'@expo/hub-android-utils': minor
```
krystofwoldrich: Only the public package can have changesets, otherwise out release ci fails.

.changeset/emulator-camera-feeds.md
```diff
@@ -0,0 +1,8 @@
...
+---
+
+Add camera image passthrough for Android emulators. Every emulator the Hub boots starts with serve-emu's camera feeds attached, and the inspector's new Camera section replaces or resets the picture per facing.
```
krystofwoldrich: I would make this short.

````suggestion
Add camera image passthrough for Android emulators.
````

krystofwoldrich (approved):
Looks great, thanks for adding the camera. Before merging pls make sure the serve-emu changes are merged and update the submodule to the latest serve-emu expo branch commit

.changeset/emulator-camera-feeds.md
vonovak: Fixed in 1913e4b: the changeset now names only `expo-device-hub`, per RELEASING.md.

.changeset/emulator-camera-feeds.md
vonovak: Applied in 1913e4b.

PR #71, opened 2026-09-07
Fix WebRTC recovery after Android capture changes

krystofwoldrich-agent:
Changing an Android capture source or gRPC image mode can leave the existing WebRTC peer attached to the previous publisher. The preview then freezes until watchdog recovery. Renegotiate when the server confirms a new capture generation, using the pending response before waiting for its first frame to commit the UI selection.

Initial metadata establishes a baseline without restarting an in-flight connection. Failed and no-op changes preserve the peer, and committing the replacement selection does not trigger a second restart. The same logic handles externally observed capture replacements and server generation resets. Pending capture metadata is React state, and a control socket that recovers before the PUT response does not finish the switch before the replacement video renders. A canvas snapshot retains the last displayed frame across peer teardown and MediaStream replacement, then clears when the new video has a frame ready. Failed/no-op switches resume the current video, and changing devices clears the saved picture.

This is the standalone recovery fix extracted from #70, based on current upstream `main` (`87c2037`). It uses the existing capture-generation contract and can merge independently: the diff contains client logic, regression tests and their development dependencies, and a patch changeset for the published `expo-device-hub` package. It has no RGB option or submodule changes.

### Validation

- On this standalone branch, `@expo/hub-client`: **168 tests passed**, build, typecheck, and lint passed.
- Hook-level tests mount DeviceScreen with the Android client and WebRTC hooks together, covering one restart per capture or image-mode replacement, no restart on first-frame commit, socket recovery before or alongside the PUT response, blocked concurrent switches, failed/no-op changes, and polled generation resets. They also cover retaining the last frame through repeated replacements and clearing it on device changes or disable. Pure helper tests also cover initial/unknown metadata.
- Frozen-lockfile installation and changeset validation passed.
- A Chromium pixel check using generated MediaStream tracks kept the old red frame visible while the replacement stream had no frame, then showed green after its first frame arrived. Hiding the retained canvas exposed the blank surface, confirming the check detects the original gap. This validates browser presentation without an emulator.
- Prior agent-browser testing of this same recovery logic in #70, using the built standalone Hub CLI, reduced RGB→MMAP recovery from roughly 18 seconds to **2.45 seconds** and measured **2.52 seconds** for MMAP→RGB. Timing runs from the capture-change PUT request to a connected replacement peer decoding at least five frames with presentation advancing. These measurements were made on the RGB feature branch, not repeated on this extracted branch. PNG uses the same generation-based recovery but has no separate timing measurement.

### DIFF
--- a/packages/@expo/hub-client/src/__tests__/webrtc-restart.test.ts
+++ b/packages/@expo/hub-client/src/__tests__/webrtc-restart.test.ts
@@ -1,7 +1,6 @@
 import { describe, expect, test } from 'bun:test';
 
 import { androidWebRtcRestartKey } from '../android-stream-source';
-import { IDLE_STREAM_SWITCH, reduceStreamSwitch } from '../stream-switch';
 import { observeWebRtcRestartKey } from '../webrtc-restart';
 
 describe('capture generation WebRTC restart', () => {
@@ -19,11 +18,6 @@ describe('capture generation WebRTC restart', () => {
   test('restarts on successful replacement before a new frame commits the UI selection', () => {
     const displayed = { sessionGeneration: 1 };
     const pending = { sessionGeneration: 2 };
-    const requesting = reduceStreamSwitch(IDLE_STREAM_SWITCH, {
-      type: 'request-start', live: true,
-    });
-    const switching = reduceStreamSwitch(requesting, { type: 'request-success', replaced: true });
-    expect(switching.phase).toBe('awaiting-interruption');
 
     const restarted = observeWebRtcRestartKey(
       { key: 1, generation: 0 },
@@ -32,17 +26,14 @@ describe('capture generation WebRTC restart', () => {
     expect(restarted).toEqual({ key: 2, generation: 1 });
 
     // Once video paints, committing pending state must not restart the new peer again.
-    expect(observeWebRtcRestartKey(restarted, androidWebRtcRestartKey(pending, null))).toBe(restarted);
+    expect(observeWebRtcRestartKey(restarted, androidWebRtcRestartKey(pending, null))).toBe(
+      restarted,
+    );
   });
 
-  test('keeps the live peer during a request and after failed or no-op replacements', () => {
+  test('preserves the restart generation when the source generation is unchanged', () => {
     const displayed = { sessionGeneration: 1 };
     const live = { key: 1, generation: 0 };
-    const requesting = reduceStreamSwitch(IDLE_STREAM_SWITCH, {
-      type: 'request-start', live: true,
-    });
-    expect(observeWebRtcRestartKey(live, androidWebRtcRestartKey(displayed, null))).toBe(live);
-    expect(reduceStreamSwitch(requesting, { type: 'request-failure' })).toBe(IDLE_STREAM_SWITCH);
     expect(observeWebRtcRestartKey(live, androidWebRtcRestartKey(displayed, null))).toBe(live);
     expect(observeWebRtcRestartKey(live, androidWebRtcRestartKey(displayed, displayed))).toBe(live);
   });
--- a/packages/@expo/hub-client/src/__tests__/webrtc-restart.test.ts
+++ b/packages/@expo/hub-client/src/__tests__/webrtc-restart.test.ts
@@ -1,0 +1,65 @@
+import { describe, expect, test } from 'bun:test';
+
+import { androidWebRtcRestartKey } from '../android-stream-source';
+import { IDLE_STREAM_SWITCH, reduceStreamSwitch } from '../stream-switch';
+import { observeWebRtcRestartKey } from '../webrtc-restart';
+
+describe('capture generation WebRTC restart', () => {
+  test('uses initial metadata as a baseline without restarting a connecting peer', () => {
+    const unknown = { key: null, generation: 0 };
+    expect(androidWebRtcRestartKey(null, null)).toBeNull();
+    const baseline = observeWebRtcRestartKey(
+      unknown,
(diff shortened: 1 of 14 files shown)


krystofwoldrich (commented):

**Verdict:** REQUEST_CHANGES (posted as a comment because the reviewer account is the PR author)
**Findings:** 0 critical · 1 design · 2 suggestion · 1 nit

The change adds a `restartKey` to `useWebRtcStream` and derives it from the Android capture `sessionGeneration`. A generation change bumps an internal counter that is a dependency of the peer-connection effect, so the peer is torn down and renegotiated. The pending PUT response feeds the key before the first frame commits the UI selection, and the commit does not restart again because the key is unchanged. The approach fits the existing switch tracker and the serve-emu generation contract. Verified locally on this branch: 158 tests pass, `tsc --noEmit` and `oxlint` pass.

The main concern is that the restart key is computed from a ref read during render, so the fix only works because `dispatchStreamSwitch` happens to trigger a re-render right after the ref is set. Details are in the inline comments.

### Not covered by an inline comment

**[suggestion] Missing changeset.** Every prior user-visible change to `@expo/hub-client` ships a `.changeset/*.md` (for example #60 `tidy-mmap-cadence.md`, #56 `grpc-scrcpy-input.md`). This is a user-facing fix in a released package and should get a `patch` changeset so it appears in the changelog.

packages/@expo/hub-client/src/useAndroidDevice.ts
```diff
@@ -829,6 +830,7 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
     sendIceServersInOffer: false,
     allowCodecFallback: false,
     onKeyframeNeeded: requestWebRtcKeyframe,
+    restartKey: androidWebRtcRestartKey(streamSource, pendingStreamSourceRef.current),
```
krystofwoldrich: **[design]** `pendingStreamSourceRef.current` is read during render. The ref is mutated in the PUT `.then` handler, and the restart only happens because `dispatchStreamSwitch({ type: 'request-success' })` always changes `streamSwitch` state right after the mutation and forces a re-render. That coupling is implicit: if the reducer ever returns the same state object for a success (or the ref is set on a path that does not dispatch), the peer keeps the frozen publisher and this fix silently stops working. React also documents refs as not to be read during render, and `eslint-plugin-react-hooks` v6 flags this pattern (`react-hooks/refs`); oxlint does not, so lint passing is not evidence here.

Suggestion: hold the pending source in React state as well (or compute the restart key inside `commitPendingStreamSource` / the success handler and store it with `useState`), so the key is derived from state the renderer owns rather than from a ref the render happens to observe.

packages/@expo/hub-client/src/useAndroidDevice.ts
```diff
@@ -866,9 +868,8 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
       setStatus('streaming');
       setError(null);
     } else if (webRtcWasLive && !webRtcGraceExpired) {
-      // When serve-emu swaps the capture source the RTP video usually keeps
-      // flowing; only the control socket is closed and reopened. Keep the frame
-      // and report a reconnect instead of an error while that settles.
+      // A capture generation change renegotiates video and reopens control.
```
krystofwoldrich, line 894: **[suggestion]** The `recovered` path of the switch tracker now behaves differently. serve-emu closes the viewer sockets before answering the PUT, so the control socket can drop and reconnect before the HTTP response arrives. In that case `reduceStreamSwitch` returns `IDLE_STREAM_SWITCH` on `request-success` (state.recovered), `commitPendingStreamSource` runs, and only then the key change restarts the peer. The sidebar reports the switch as finished and `streamSourcePending` re-enables the controls while the frame goes to `reconnecting` for the renegotiation, and a second PUT is accepted mid-renegotiation.

Before this PR the recovered path meant the stream really was back. Consider treating a `replaced` success as `AWAITING_FRAME` when the transport is WebRTC, since the client knows it is about to restart the peer, or gating `recovered` on `webRtcLive` rather than on the control socket.

packages/@expo/hub-client/src/__tests__/webrtc-restart.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      androidWebRtcRestartKey({ sessionGeneration: 1 }, null),
+    );
+    expect(baseline).toEqual({ key: 1, generation: 0 });
+    expect(observeWebRtcRestartKey(baseline, 1)).toBe(baseline);
+  });
+
+  test('restarts on successful replacement before a new frame commits the UI selection', () => {
+    const displayed = { sessionGeneration: 1 };
+    const pending = { sessionGeneration: 2 };
+    const requesting = reduceStreamSwitch(IDLE_STREAM_SWITCH, {
```
krystofwoldrich: **[nit]** These `reduceStreamSwitch` assertions duplicate `stream-switch.test.ts` and do not exercise the part of this change that carries the risk: the render-time key derivation in `useAndroidDevice` and the effect dependency in `useWebRtcStream`. The pure helpers are well covered, but a hook-level test (the package already has a `.test.tsx` with React rendering) that asserts one restart per generation change and none on commit would protect the wiring that the inline design comment is about.

packages/@expo/hub-client/src/webrtc-restart.ts
```diff
@@ -0,0 +1,18 @@
+export type WebRtcRestartKey = string | number | null;
```
krystofwoldrich: **[nit]** `string` is not used by any caller; the Android key is always `number | null`. Narrowing the type to what is used keeps the contract honest until a string key exists.

PR #73, opened 2026-09-07
Fix rotation and touch alignment for gRPC screens

krystofwoldrich-agent:
Fix rotation and touch alignment for gRPC screens. Selecting portrait or landscape synchronizes the emulator accelerometer with Android's rotation lock, and direct gRPC touches map back to the native framebuffer.

Apps such as the launcher can keep Android's display portrait while gRPC produces a landscape screenshot. Scrcpy input now uses the difference between screenshot rotation and Android's active display rotation to transform touch coordinates and dimensions. It reads display rotation at each gesture's start, catching app switches even when video dimensions stay unchanged, and keeps one mapping throughout a multi-pointer drag. Rotation-read failures reject the affected touch while leaving subsequent controls usable.

Includes the implementation from [serve-emu #25](https://github.com/expo/serve-emu/pull/25) directly in the monorepo's `packages/serve-emu` source, rebased onto upstream `main` at `3d13723`. Includes documentation and a patch changeset.

### Validation

- On the rebased branch, all 1,072 backend tests passed. The backend aggregate check also passed coverage, documentation synchronization, typechecks, builds, and package smoke tests.
- The launcher regression failed in four mismatched-orientation cases before the fix; all eight orientation cases now pass. Additional integration coverage checks app switches without video resizing, multi-pointer drags, swipe endpoints, read-failure recovery, and hardware controls without rotation queries.
- On the rebuilt CLI with the default RGB888 gRPC stream and scrcpy input, verified the launcher reports display rotation 0 while screenshots report rotation 1. A swipe along the sideways content's top edge opens the notification shade correctly.
- Verified a launcher-icon tap opens Chrome, the mapping changes from relative rotation 1 to 0 with unchanged screenshot dimensions, and a tap on Chrome's address bar focuses it after the app transition finishes.

### DIFF
--- a/packages/serve-emu/packages/serve-emu/src/adb.ts
+++ b/packages/serve-emu/packages/serve-emu/src/adb.ts
@@ -1,4 +1,5 @@
 import { spawn } from "node:child_process";
+import { isEmulatorSerial } from "./device-capabilities.ts";
 import { execBuffer, execText, type ExecResult } from "./exec.ts";
 
 const ADB_QUERY_TIMEOUT_MS = 2_000;
@@ -264,6 +265,24 @@ export async function setUserRotation(
     throw new Error(
       `adb shell ${args.join(" ")} failed: ${execFailure(r)}`,
     );
+  }
+  if (isEmulatorSerial(serial) && orientation !== "auto") {
+    // gRPC screenshots follow the emulator's accelerometer, independently of
+    // Android's user-rotation lock. Keep both in the same orientation so the
+    // streamed image is upright and its dimensions follow the Rotate control.
+    const acceleration = orientation === "portrait" ? "0:9.8:0" : "9.8:0:0";
+    const sensor = await runExec(
+      "adb",
+      ["-s", serial, "emu", "sensor", "set", "acceleration", acceleration],
+      { timeout: ADB_MUTATION_TIMEOUT_MS },
+    );
+    // The emulator console can report KO while adb exits successfully.
+    const output = `${sensor.stdout}\n${sensor.stderr}`.trim();
+    if (execFailed(sensor) || /^KO\b/m.test(output)) {
+      throw new Error(
+        `adb emu sensor set acceleration failed: ${execFailure(sensor)}`,
+      );
+    }
   }
   return getUserRotation(serial, runExec);
 }
--- a/packages/serve-emu/packages/serve-emu/src/grpc-session.ts
+++ b/packages/serve-emu/packages/serve-emu/src/grpc-session.ts
@@ -1,4 +1,5 @@
 import { setTimeout as delay } from "node:timers/promises";
+import { getDisplayRotation, type DisplayRotation } from "./adb.ts";
 import {
   ControlInputQueue,
   ControlInputRejectedError,
@@ -41,6 +42,7 @@ import {
   compileGesture,
   normalizeTextForControl,
   type Gesture,
+  type Screen,
 } from "./input.ts";
 import {
   SCRCPY_DEFAULTS,
@@ -976,6 +978,28 @@ function createGrpcImageCaptureTransport(options: {
   };
 }
 
+function rotateUnitTouch(x: number, y: number, rotation: number) {
+  if (rotation === 1) return { x: 1 - y, y: x };
+  if (rotation === 2) return { x: 1 - x, y: 1 - y };
+  if (rotation === 3) return { x: y, y: 1 - x };
+  return { x, y };
+}
+
+function rotateScrcpyGesture(gesture: Gesture, rotation: number): Gesture {
+  switch (gesture.type) {
+    case "tap":
+    case "touch":
+      return { ...gesture, ...rotateUnitTouch(gesture.x, gesture.y, rotation) };
+    case "swipe": {
+      const start = rotateUnitTouch(gesture.x1, gesture.y1, rotation);
+      const end = rotateUnitTouch(gesture.x2, gesture.y2, rotation);
+      return { ...gesture, x1: start.x, y1: start.y, x2: end.x, y2: end.y };
+    }
+    default:
+      return gesture;
+  }
+}
+
 export type GrpcDisplayGeometry = {
   encodedSize: { width: number; height: number };
   touchSize: { width: number; height: number };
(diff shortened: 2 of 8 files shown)


packages/serve-emu/packages/serve-emu/src/grpc-session.ts
```diff
@@ -2238,12 +2271,50 @@ export async function startGrpcSession(
 
   const createScrcpyControls = (session: ScrcpyControlSession) => {
     const writer = new SocketControlWriter(session.controlSocket);
+    const readRotation = dependencies.readDisplayRotation ??
+      ((serial: string, signal: AbortSignal) => getDisplayRotation(serial, undefined, signal));
```
krystofwoldrich-agent, line 2275: **[design]** `getDisplayRotation` runs `adb shell dumpsys window displays` with `lane: "background"` and a 2 s timeout. `readGeometry` awaits it for every tap and swipe and for the first `down` of each touch gesture, so every scrcpy gesture now waits on an ADB round trip before its first packet. The background lane is the lowest priority in `exec.ts`: `#canStartLane` reserves active slots for interactive work and `#drain` starts interactive and default jobs first, so a tap can also queue behind display-size polls. The package `AGENTS.md` asks to optimize input for low latency.

Suggested direction: cache the display rotation, refresh it from the existing `GrpcNativeTouchGeometryMonitor` poll and on image geometry changes, and only read synchronously when the cache is missing. If the per-gesture read stays, use `lane: "interactive"` like the control-path ADB calls in `accessibility.ts` and `location.ts`.

packages/serve-emu/packages/serve-emu/src/adb.ts
```diff
@@ -265,6 +266,24 @@ export async function setUserRotation(
...
+    // streamed image is upright and its dimensions follow the Rotate control.
+    const acceleration = orientation === "portrait" ? "0:9.8:0" : "9.8:0:0";
+    const sensor = await runExec(
+      "adb",
+      ["-s", serial, "emu", "sensor", "set", "acceleration", acceleration],
+      { timeout: ADB_MUTATION_TIMEOUT_MS },
+    );
+    // The emulator console can report KO while adb exits successfully.
+    const output = `${sensor.stdout}\n${sensor.stderr}`.trim();
+    if (execFailed(sensor) || /^KO\b/m.test(output)) {
```
krystofwoldrich-agent, line 281: **[suggestion]** This is the third emulator-console `KO` convention: `readEmuConsoleValue` above filters `KO:` lines and returns null, and `assertGeoFixOutput` in `location.ts` checks `/^KO\b/` on the combined output. A shared `runEmuConsoleCommand(serial, args, runExec)` that throws on a failed exit or a `KO` line would let this and `location.ts` share one implementation.

packages/serve-emu/packages/serve-emu/src/adb.ts
```diff
@@ -265,6 +266,24 @@ export async function setUserRotation(
...
+      throw new Error(
```
krystofwoldrich-agent, line 282: **[suggestion]** At this point the guest rotation lock already succeeded, so the device is left with Android locked to the new rotation and the accelerometer still in the old pose. The error does not say that. Consider a message such as "Android rotation lock applied, but the emulator sensor update failed" so the UI and agents can tell this apart from a failed lock.

packages/serve-emu/packages/serve-emu/src/grpc-session.ts
```diff
@@ -2238,12 +2271,50 @@ export async function startGrpcSession(
...
+          let screen = nativeTouchSize;
+          let mapped = gesture;
+          if (gesture.type === "tap" || gesture.type === "swipe") {
+            const geometry = await readGeometry(signal);
+            screen = geometry.screen;
+            mapped = rotateScrcpyGesture(gesture, geometry.rotation);
+          } else if (gesture.type === "touch") {
+            const pointerId = gesture.pointerId ?? 0;
+            // A failed down must not leave subsequent moves using stale geometry.
+            if (gesture.action !== "down" && !pointers.has(pointerId)) return;
```
krystofwoldrich-agent, line 2306: **[nit]** This early `return` resolves the gesture's `completion` as success without writing a packet. Before this PR such moves and ups were forwarded to scrcpy. Dropping them after a failed `down` is reasonable, but a `move` without any prior `down` now becomes a silent no-op. A `ControlInputRejectedError` or a note in the gesture API docs would make the contract explicit.

krystofwoldrich-agent (commented):

**Verdict:** REQUEST_CHANGES (posted as a comment because the reviewing account authored this PR)
**Findings:** 0 critical · 1 design · 2 suggestion · 1 nit

The change has two parts. `setUserRotation` now also sets the emulator accelerometer so gRPC screenshots follow the Rotate control. gRPC touch input undoes the image rotation before scaling, and scrcpy input maps image space to Android display space using `image.rotation - displayRotation`. The rotation math composes correctly (`rotateUnitTouch` forms a consistent group, and the unit tests cover all 4x2 orientation pairs), and `nativeTouchSize` already tracks the oriented image size, so swapping it on odd rotations yields the unrotated framebuffer. The main concern is latency: every scrcpy tap, swipe, and first touch-down now waits on an `adb shell dumpsys window displays` round trip that runs in the lowest-priority exec lane.

### Findings

#### [design] Per-gesture `dumpsys` in the background lane adds ADB latency to every scrcpy tap
`packages/serve-emu/packages/serve-emu/src/grpc-session.ts:2275`
`getDisplayRotation` is a query helper and runs with `lane: "background"` and a 2 s timeout. `readGeometry` calls it for every tap and swipe and for the first `down` of each touch gesture, so the gesture waits for a full `dumpsys window displays` before the first packet is written. The background lane is the lowest priority: `#canStartLane` keeps active slots in reserve for interactive work and `#drain` starts interactive and default jobs first, so a tap can also queue behind display-size polls or other background ADB calls. `AGENTS.md` for this package asks to optimize for low latency and forbids `adb shell input` for exactly this reason. Suggested direction: cache the display rotation and refresh it from the existing `GrpcNativeTouchGeometryMonitor` poll and on image geometry changes, and only read synchronously when the cache is missing. If a per-gesture read is kept, run it with `lane: "interactive"` like the other control-path ADB calls in `accessibility.ts` and `location.ts`.

#### [suggestion] Third emulator-console KO convention in the codebase
`packages/serve-emu/packages/serve-emu/src/adb.ts:281`
`readEmuConsoleValue` in this file filters `KO:` lines and returns null, `assertGeoFixOutput` in `location.ts` checks `/^KO\b/` on the combined output, and this hunk adds a third inline variant with the `m` flag and its own error message. A small shared helper such as `runEmuConsoleCommand(serial, args, runExec)` that throws on a failed exit or a `KO` line would let `setUserRotation` and `location.ts` share one implementation.

#### [suggestion] Error after a successful lock leaves a mixed state the caller cannot see
`packages/serve-emu/packages/serve-emu/src/adb.ts:282`
When the sensor command fails, the guest rotation lock has already been applied but the accelerometer still points the old way. This is the mismatched state the PR fixes, and the thrown error does not say the lock succeeded. Consider stating that in the message (for example "Android rotation lock applied, but the emulator sensor update failed") so the UI and agents can tell the two cases apart.

#### [nit] Moves and ups for unknown pointers complete successfully without sending anything
`packages/serve-emu/packages/serve-emu/src/grpc-session.ts:2306`
The early `return` resolves the queued gesture's `completion` as success even though no packet reached scrcpy. Before this PR those events were forwarded. Dropping them is reasonable after a failed `down`, but a caller that sends a `move` without a prior `down` now gets a silent no-op. A `ControlInputRejectedError` or a comment in the API docs would make the contract explicit.

PR #81, opened 2026-09-09
Fix delayed RGB888 playback at lower video FPS

krystofwoldrich:
Fixes delayed RGB888 playback when the emulator produces frames faster than the configured video FPS. The previous implementation paused the gRPC reader at the encoder rate, allowing old frames to accumulate upstream. Update serve-emu to continuously receive RGB888 responses, keep only the latest image, and apply `--video-fps` to fresh encoder submissions.

Stacked on https://github.com/expo/expo-device-hub/pull/80 using `gh stack`. Includes the capture fix and real HTTP/2 regression test from https://github.com/expo/serve-emu/pull/28 directly in the monorepo under `packages/serve-emu` and corrects the pacing documentation. No serve-emu submodule update or separate capture PR is required.

Capture statistics distinguish RGB888 decoded responses from encoder selection. The documentation explains why predecode coalescing remains zero and how usable-image and fresh-encoder rates expose local frame selection. HTTP/2 integration tests share response-framing fixtures and report explicit deadline failures.

Validation:

- `bun run build`, `bun run typecheck`, `bun run lint`, and `bun run test` passed in the monorepo: 1,478 tests, including 932 serve-emu tests and 546 Device Hub/UI tests.
- `bun run --filter serve-emu check` passed on the monorepo layout: 932 tests, coverage, typechecks, builds, documentation checks, and package smoke test.
- A real HTTP/2 regression verifies that a bounded burst of 121 RGB888 responses reaches the newest encoded image promptly while the encoder is configured for 30 FPS and submits fewer frames. Existing encoder-backpressure, rotation, cancellation, PNG, and MMAP tests pass.
- Live WebRTC validation at 30 FPS, `-gpu host`, 540×1170 native images, and the same animated workload: source timestamp → gRPC reception p50 decreased from 5179.5 ms to 14.7 ms; p95 decreased from 5537.4 ms to 126.1 ms. Incoming rates were 27.54 and 27.20 responses/sec; fresh encoder submission rates were 27.54 and 19.61/sec. These capture-boundary measurements show the removal of accumulated delay and are not a general throughput or end-to-end latency guarantee.

Authored with Codex (GPT-6).

### DIFF
--- a/packages/serve-emu/README.md
+++ b/packages/serve-emu/README.md
@@ -640,9 +640,13 @@ rotation or display resizing.
 
 This path still requires GPU-to-CPU readback inside the emulator. It is not GPU
 texture sharing, zero-copy, or hardware encoding; speed depends on the workload
-and must be measured. A bounded gRPC message pacer and a single latest image
-preserve encoder backpressure. `--max-fps` paces local consumption and encoder
-submission; it does not impose a server-side screenshot FPS limit. Health
+and must be measured. RGB888 drains incoming gRPC responses continuously and
+retains only the latest image for the encoder. `--max-fps` limits fresh encoder
+submissions independently of incoming frame rate: excess images are replaced,
+so lowering encoded FPS does not queue old source frames for delayed playback.
+Encoder backpressure also retries the latest image. PNG retains its predecode
+message pacer; MMAP selects notifications before reading shared memory. None of
+these local limits impose a server-side screenshot FPS limit. Health
 reports `grpcCapture.imageMode`, actual received `grpcMessageBytesReceived`,
 protobuf decode timing, and separate received/source/encoder frame rates.
 MMAP counters remain zero and shared-memory timing remains null for RGB888.
--- a/packages/serve-emu/README.md
+++ b/packages/serve-emu/README.md
@@ -651,6 +651,13 @@ reports `grpcCapture.imageMode`, actual received `grpcMessageBytesReceived`,
 protobuf decode timing, and separate received/source/encoder frame rates.
 MMAP counters remain zero and shared-memory timing remains null for RGB888.
 
+For RGB888, `rawGrpcMessagesEmitted` counts responses passed to protobuf decoding
+and matches `rawGrpcMessagesReceived`; `rawGrpcMessagesCoalesced` stays zero.
+Those message counters exclude images replaced in the latest-image slot before
+encoding. Compare `usableImageFps` with `freshEncoderWriteFps` (which excludes
+repeats) to observe the local reduction before encoder submission. Their
+rolling-rate difference is not an exact cumulative dropped-image count.
+
 ```sh
 serve-emu -s emulator-5554 --stream-mode grpc-screenshot --grpc-image-mode rgb888
 curl -X PUT http://localhost:3300/api/stream-mode \
--- a/packages/serve-emu/packages/serve-emu/src/grpc-session.ts
+++ b/packages/serve-emu/packages/serve-emu/src/grpc-session.ts
@@ -693,9 +693,11 @@ export function grpcImageModeBehavior(
   }
   return {
     encoderInputFormat: "rgb24",
-    // In-band RGB messages own their pixels and can use the same bounded
-    // predecode pacer as PNG. MMAP metadata must be decoded immediately.
-    predecodeMaxFps: imageMode === "rgb888" ? maxFps : undefined,
+    // Drain raw RGB responses independently of encoder FPS. Pausing HTTP/2
+    // queues old pixels upstream; onImage keeps only the newest frame and the
+    // encoder write pacer decides when to submit it. MMAP also drains metadata
+    // immediately, then selects which notifications trigger memory reads.
+    predecodeMaxFps: undefined,
     needsEncoderFollowUp: (repeat) => !repeat,
   };
 }
--- a/packages/serve-emu/packages/serve-emu/tests/grpc-session.test.ts
+++ b/packages/serve-emu/packages/serve-emu/tests/grpc-session.test.ts
@@ -29,7 +29,6 @@ import {
 } from "../src/grpc-session.ts";
 import {
   EmulatorGrpcClient,
-  encodeImageFormat,
   IMG_FORMAT_PNG,
   IMG_FORMAT_RGB888,
   type EmuImage,
@@ -43,6 +42,7 @@ import type {
   ScrcpyControlSession,
   VideoFrame,
 } from "../src/scrcpy.ts";
+import { encodeEmulatorImage, grpcFrame } from "./fixtures/grpc.ts";
 
 const CONFIG_FRAME: VideoFrame = {
   type: "frame",
@@ -1189,27 +1189,18 @@ describe("startGrpcSession integration", () => {
     const width = 192;
     const height = 192;
     const frameCount = 120;
-    const varint = (value: number): Buffer => {
-      const bytes: number[] = [];
-      do {
-        const byte = value & 0x7f;
-        value >>>= 7;
-        bytes.push(value ? byte | 0x80 : byte);
-      } while (value);
-      return Buffer.from(bytes);
-    };
-    const format = encodeImageFormat({ format: IMG_FORMAT_RGB888, width, height });
     const frame = (sequence: number): Buffer => {
       const pixels = Buffer.alloc(width * height * 3);
       pixels.writeUInt32LE(sequence);
-      const body = Buffer.concat([
-        Buffer.from([0x0a]), varint(format.length), format,
-        Buffer.from([0x22]), varint(pixels.length), pixels,
-        Buffer.from([0x28]), varint(sequence),
-      ]);
-      const header = Buffer.alloc(5);
-      header.writeUInt32BE(body.length, 1);
-      return Buffer.concat([header, body]);
+      return grpcFrame(
+        encodeEmulatorImage({
+          format: IMG_FORMAT_RGB888,
+          width,
+          height,
+          image: pixels,
+          seq: sequence,
+        }),
+      );
     };
     let screenshotStream: ServerHttp2Stream | undefined;
     const server = http2.createServer();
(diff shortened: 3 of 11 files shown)


packages/serve-emu/README.md
```diff
@@ -640,9 +640,13 @@ rotation or display resizing.
...
 This path still requires GPU-to-CPU readback inside the emulator. It is not GPU
 texture sharing, zero-copy, or hardware encoding; speed depends on the workload
-and must be measured. A bounded gRPC message pacer and a single latest image
-preserve encoder backpressure. `--max-fps` paces local consumption and encoder
-submission; it does not impose a server-side screenshot FPS limit. Health
+and must be measured. RGB888 drains incoming gRPC responses continuously and
+retains only the latest image for the encoder. `--max-fps` limits fresh encoder
+submissions independently of incoming frame rate: excess images are replaced,
+so lowering encoded FPS does not queue old source frames for delayed playback.
+Encoder backpressure also retries the latest image. PNG retains its predecode
```
krystofwoldrich-agent, line 647: **[suggestion]** The PR diagnoses `GrpcMessagePacer.pause()` as the cause of stale playback: pausing the HTTP/2 readable lets the emulator's frames queue upstream in FIFO order. PNG mode is still selectable via `--grpc-image-mode png` and still uses that pacer, so it keeps the same delayed-playback behavior when the emulator outpaces `--max-fps`.

A drain-and-coalesce pacer (never pause, keep only the newest raw message, decode once per slot) would fix PNG as well and would keep `rawGrpcMessagesCoalesced` meaningful. Fine as a follow-up; noting it here because the README now presents the PNG pacer as intentional.

packages/serve-emu/packages/serve-emu/src/grpc-session.ts
```diff
@@ -693,9 +693,11 @@ export function grpcImageModeBehavior(
...
   return {
     encoderInputFormat: "rgb24",
-    // In-band RGB messages own their pixels and can use the same bounded
-    // predecode pacer as PNG. MMAP metadata must be decoded immediately.
-    predecodeMaxFps: imageMode === "rgb888" ? maxFps : undefined,
+    // Drain raw RGB responses independently of encoder FPS. Pausing HTTP/2
+    // queues old pixels upstream; onImage keeps only the newest frame and the
+    // encoder write pacer decides when to submit it. MMAP also drains metadata
+    // immediately, then selects which notifications trigger memory reads.
+    predecodeMaxFps: undefined,
```
krystofwoldrich-agent, line 700: **[suggestion]** With `predecodeMaxFps` unset, `#request` uses the plain `GrpcMessageParser` path (emulator-grpc.ts ~line 995), which emits both `received` and `emitted` for every message. For RGB888, `rawGrpcMessagesEmitted` now always equals `rawGrpcMessagesReceived` and `rawGrpcMessagesCoalesced` stays at 0.

The Hub dashboard shows these as "Selected notifications" and "Coalesced notifications" for every gRPC mode (`StreamStatistics.tsx:323-327`). Frames are still dropped, now by `latest` replacement in `onImage`, but nothing counts that. In the PR's own measurement (27.2 incoming vs 19.6 fresh submissions/sec) the dashboard would show 0 coalesced.

Options: count images replaced before an encoder write in `GrpcCaptureDiagnosticsTracker`, or document that for RGB888 the drop rate is visible only as "Usable image FPS" minus "Encoder input FPS".

packages/serve-emu/packages/serve-emu/tests/grpc-session.test.ts
```diff
@@ -1182,6 +1185,88 @@ async function waitFor(predicate: () => boolean): Promise<void> {
 }
 
 describe("startGrpcSession integration", () => {
+  test("RGB888 drains incoming frames independently of the encoder FPS limit", async () => {
+    const width = 192;
+    const height = 192;
+    const frameCount = 120;
+    const varint = (value: number): Buffer => {
```
krystofwoldrich-agent: **[nit]** `tests/emulator-grpc.test.ts` already defines `grpcFrame` for the 5-byte gRPC header. This is now the second real HTTP/2 integration test, so a shared test helper (frame header plus an `EmulatorImage` protobuf encoder) would remove this hand-rolled varint and the byte-literal image bodies in the other file.

packages/serve-emu/packages/serve-emu/tests/grpc-session.test.ts
```diff
@@ -1182,6 +1185,88 @@ async function waitFor(predicate: () => boolean): Promise<void> {
...
+        screenshotStream!.write(frame(sequence));
+      }
+      const deadline = performance.now() + 2000;
+      while (performance.now() < deadline && encoders[0]!.lastBuffer!.readUInt32LE() !== frameCount) {
+        await new Promise((resolve) => setTimeout(resolve, 5));
+      }
+      expect(session.diagnostics!().grpcCapture!.rawGrpcMessagesReceived).toBe(frameCount + 1);
+      expect(encoders[0]!.lastBuffer!.readUInt32LE()).toBe(frameCount);
+      expect(encoders[0]!.options).toMatchObject({ fps: 30, inputFormat: "rgb24" });
+      expect(encoders[0]!.writes).toBeLessThan(frameCount / 2);
```
krystofwoldrich-agent, line 1262: **[nit]** The polling loop exits silently on the 2 s deadline, so a slow CI run fails on `toBe(frameCount)` with a confusing frame index instead of a timeout message. Throwing on deadline expiry inside the loop (like `waitFor` does) would make the failure reason clear.

The `writes < frameCount / 2` bound itself is consistent: `boundaryDelayMs()` is at least 1.5× the frame interval, so follow-up repeats are superseded during a continuous burst and writes track elapsed time at 30 FPS.

krystofwoldrich-agent (commented):

## Review: Fix delayed RGB888 playback at lower video FPS

**Verdict:** COMMENT
**Findings:** 0 critical · 0 design · 2 suggestion · 2 nit

The change sets `predecodeMaxFps` to `undefined` for RGB888. The client then uses the plain `GrpcMessageParser` instead of `GrpcMessagePacer`, so the HTTP/2 stream is never paused. Every response is decoded as a view, `onImage` replaces `latest`, and `GrpcFrameWritePacer` limits fresh encoder writes to `--max-fps`. I verified the write loop: a fresh accepted write resets the follow-up timer, so a fast source yields exactly one fresh write per slot. Encoder rejections retry through `scheduleWrite`, which matches the README claim. Both serve-emu READMEs stay in sync (`docs:check` passes) and `grpc-session.test.ts` passes locally (42 tests). The new HTTP/2 regression test fails against the old pacer, so it protects the fix.

The diagnosis is correct and the fix is minimal. The two suggestions are about what the fix leaves behind: the same pause-based pacer still serves PNG mode, and two dashboard counters lose their meaning for RGB888.

### Findings

#### [suggestion] PNG mode keeps the same upstream-queuing root cause
`packages/serve-emu/README.md:647`
The PR diagnoses `GrpcMessagePacer.pause()` as the cause of stale playback. PNG is still a selectable `--grpc-image-mode` and still uses that pacer. A drain-and-coalesce pacer (keep reading, keep only the newest raw message, decode once per slot) would fix PNG too and keep the `coalesced` counter meaningful. Fine as a follow-up.

#### [suggestion] "Selected" and "Coalesced notifications" become uninformative for RGB888
`packages/serve-emu/packages/serve-emu/src/grpc-session.ts:700`
The plain parser emits `received` and `emitted` for every message, so `rawGrpcMessagesEmitted` equals `rawGrpcMessagesReceived` and `rawGrpcMessagesCoalesced` stays at 0. The dashboard shows both for every gRPC mode. Frames are still dropped, now by `latest` replacement in the session, and that is not counted. Consider counting replaced images or documenting that RGB888 drop rate is `usableImageFps - encoderInputFps`.

#### [nit] Test duplicates the gRPC frame helper
`packages/serve-emu/packages/serve-emu/tests/grpc-session.test.ts:1192`
`tests/emulator-grpc.test.ts` already has `grpcFrame`. This is the second real HTTP/2 test; a shared helper for the frame header and an `EmulatorImage` protobuf encoder would remove the hand-rolled varint.

#### [nit] Deadline loop hides the failure reason
`packages/serve-emu/packages/serve-emu/tests/grpc-session.test.ts:1262`
If the burst is slower than 2 s, the assertions report a wrong frame index rather than a timeout. Failing inside the loop on deadline expiry would make CI failures easier to read.

packages/serve-emu/packages/serve-emu/src/grpc-session.ts
krystofwoldrich: Addressed in https://github.com/expo/expo-device-hub/commit/cefa2c4f8c4aade40a5183e484c3a979d79453d1 using the documentation approach. RGB888 now labels these counters “Decoded responses” and “Predecode coalescing”. The API comments and both READMEs explain that these counters exclude images replaced before encoding, and that comparing Usable image FPS with Encoder input FPS shows the local reduction in fresh submissions, not an exact cumulative drop count. Updated the dashboard fixture and HTTP/2 regression to verify that every response is decoded and predecode coalescing stays zero while fewer frames reach the encoder.

Validation: focused tests, full repository checks, and GitHub CI passed.

— Codex (GPT-6)

packages/serve-emu/packages/serve-emu/tests/grpc-session.test.ts
krystofwoldrich: Addressed in https://github.com/expo/expo-device-hub/commit/cefa2c4f8c4aade40a5183e484c3a979d79453d1. Added shared `grpcFrame` and `encodeEmulatorImage` helpers in `tests/fixtures/grpc.ts`. Both HTTP/2 integration test files now use them, replacing the duplicated framing/varint code and literal response bodies. The protocol unit tests retain their independent byte-level expectations.

Validation: focused tests, full repository checks, and GitHub CI passed.

— Codex (GPT-6)

packages/serve-emu/packages/serve-emu/tests/grpc-session.test.ts
krystofwoldrich: Addressed in https://github.com/expo/expo-device-hub/commit/cefa2c4f8c4aade40a5183e484c3a979d79453d1. The polling loop now throws explicitly when the 2000 ms deadline expires, reporting the target frame and last frame received by the encoder. This makes a timeout distinguishable from a frame-value assertion failure; the encoder-write bound remains unchanged.

Validation: focused tests, full repository checks, and GitHub CI passed.

— Codex (GPT-6)

PR #84, opened 2026-09-09
Restart Android WebRTC promptly when stream settings change

krystofwoldrich:
Changing Android stream settings could bring back a green **Live** label while WebRTC still displayed a frozen frame. Source switches reopened the input socket without replacing the dead video peer; the browser then had to detect ICE loss and wait through a 10-second grace period and retry backoff. In the reproduction, the replacement peer started roughly 18 seconds after the server closed the control socket.

Restart WebRTC immediately when that socket reports a deliberate server shutdown. Resolution changes leave the socket open, so restart after their settings PATCH succeeds instead. Reuse the existing switch tracker to keep the label and stream options pending until fresh video arrives. Save the last decoded frame as the video poster before closing the peer, so replacement does not flash black. Failed or superseded settings writes do not restart the peer.

This is stacked on #82. The change stays in the client hooks; it does not alter server capture staging or the existing recovery timeouts. The control socket reads the latest restart callback through a layout-synchronized ref, so video remounts do not reopen it. The existing eight-second frame wait keeps the poster visible beyond the ordinary five-second reconnect grace. A future cleanup could give encoder replacements one explicit lifecycle signal, but a larger refactor is not needed for this fix. PNG snapshot conversion optimization is deferred to a follow-up.

Validation:

- 210 client/component tests pass, including Android-client orchestration: stable control sockets across video remounts, pending guards, successful/failed/superseded settings writes, poster capture/removal, and the frame-wait timeout.
- Client type checking, lint, client build, and web export pass.
- Tested on a newly booted `Codex_WebRTC_Reconnect_API_36_1` emulator (`emulator-5556`) with an animated page on port 3499 and the Hub from this worktree on port 50240.
- Browser checks covered capture source, input source, PNG/MMAP, and resolution changes. The final recording started a new peer 21 ms after the server close, removing the previous ~18-second client wait. Total recovery still includes encoder startup and negotiation: the recorded changes took about 9 seconds under local emulator load, with approximately 5 seconds spent staging the replacement on the server.

The recording shows resolution, capture-source, and image-mode changes: **Reconnecting** appears immediately, the last frame remains visible, and **Live** returns with fresh video as the options become available.

Authored by Codex (GPT-6).

https://github.com/user-attachments/assets/ea7baee7-899f-4f80-ad9f-0b5e50694af9

### DIFF
--- a/packages/@expo/hub-client/package.json
+++ b/packages/@expo/hub-client/package.json
@@ -44,8 +44,10 @@
     "@swc/core": "^1.15.43",
     "@types/bun": "latest",
     "@types/react": "~19.2.2",
+    "@types/react-test-renderer": "19.1.0",
     "oxfmt": "^0.56.0",
     "oxlint": "^1.71.0",
+    "react-test-renderer": "19.2.3",
     "typescript": "~6.0.3"
   },
   "engines": {
--- a/packages/@expo/hub-client/src/useAndroidDevice.ts
+++ b/packages/@expo/hub-client/src/useAndroidDevice.ts
@@ -835,6 +835,12 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
     restartWebRtcStream();
   }, [restartWebRtcStream, webRtcVideoElement]);
 
+  // Media remounts update the restart target without replacing the input socket.
+  const restartWebRtcRef = useRef(restartWebRtc);
+  useLayoutEffect(() => {
+    restartWebRtcRef.current = restartWebRtc;
+  }, [restartWebRtc]);
+
   const updateStreamSettings = useCallback(
     (patch: Partial<DeviceStreamEncoderSettings>) => {
       if (streamSourceLoadingRef.current || isStreamSwitchPending(streamSwitchRef.current)) return;
@@ -850,11 +856,11 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
         }
         // Resolution changes replace the encoder without closing the input
         // socket. A new peer avoids waiting on the old decoder's video state.
-        restartWebRtc();
+        restartWebRtcRef.current();
         dispatchStreamSwitch({ type: 'request-success', replaced: true });
       });
     },
-    [dispatchStreamSwitch, restartWebRtc, useWebRtc, writeStreamSettings],
+    [dispatchStreamSwitch, useWebRtc, writeStreamSettings],
   );
 
   const webRtcLive =
@@ -891,9 +897,11 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
     if (webRtcLive) {
       setStatus('streaming');
       setError(null);
-    } else if (webRtcWasLive && !webRtcGraceExpired) {
+    } else if (
+      webRtcWasLive && (!webRtcGraceExpired || streamSwitch.phase === 'awaiting-frame')
+    ) {
       // A source switch replaces both the control socket and the video peer.
-      // Keep the last frame until the replacement peer delivers fresh video.
+      // Its bounded frame wait can outlast the ordinary reconnect grace period.
       setStatus('reconnecting');
       setError(null);
     } else if (webRtcError) {
@@ -906,7 +914,15 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
       setStatus('connecting');
       setError(null);
     }
-  }, [useWebRtc, webRtcError, webRtcGraceExpired, webRtcInputError, webRtcLive, webRtcWasLive]);
+  }, [
+    useWebRtc,
+    webRtcError,
+    webRtcGraceExpired,
+    webRtcInputError,
+    webRtcLive,
+    webRtcWasLive,
+    streamSwitch.phase,
+  ]);
 
   // Attach the negotiated MediaStream to DeviceScreen's current <video> node.
   // The node is stateful (rather than only a ref) so a remount reattaches the
@@ -1398,7 +1414,7 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
           // serve-emu stops the old video peer along with this control socket.
           // Renegotiate now instead of waiting for ICE loss and its grace period;
           // the new input socket alone must not make the old video read as live.
-          restartWebRtc();
+          restartWebRtcRef.current();
         }
         retryInput('WebRTC input disconnected. Retrying...', event.code, wasHealthy);
       };
@@ -1422,7 +1438,7 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
       if (wsRef.current === ws) wsRef.current = null;
       setWebRtcInputReady(false);
     };
-  }, [active, baseUrl, targetDevice, useWebRtc, restartWebRtc]);
+  }, [active, baseUrl, targetDevice, useWebRtc]);
 
   // ── Logcat (SSE, best-effort) — off by default; opt-in via attach ──
   useEffect(() => {
--- a/packages/@expo/hub-client/src/useAndroidDevice.ts
+++ b/packages/@expo/hub-client/src/useAndroidDevice.ts
@@ -50,6 +50,7 @@ import { KeyedWriteTracker } from './keyed-write-tracker';
 import { androidMessageForKeyboardInput } from './keyboard';
 import { MsePlayer } from './mse-player';
 import {
+  isDeliberateServerClose,
   RECONNECT_BASE_DELAY_MS,
   STREAM_RECONNECT_GRACE_MS,
   scheduleReconnect,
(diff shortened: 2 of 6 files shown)


packages/@expo/hub-client/src/useAndroidDevice.ts
```diff
@@ -1373,7 +1422,7 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
       if (wsRef.current === ws) wsRef.current = null;
       setWebRtcInputReady(false);
     };
-  }, [active, baseUrl, targetDevice, useWebRtc]);
+  }, [active, baseUrl, targetDevice, useWebRtc, restartWebRtc]);
```
krystofwoldrich-agent: **[design]** The control socket now reconnects whenever the video element changes.

`restartWebRtc` depends on `webRtcVideoElement`, and this effect now depends on `restartWebRtc`. `webRtcVideoElement` starts as null and is set by the `attachVideo` ref callback after the first commit. So when WebRTC turns on, this effect runs once with a null element. The re-render then changes `restartWebRtc`, and the effect closes that socket and opens a second one. The same happens on any later remount of the video node.

Each cycle sets `webRtcInputReady` to false. That flips the status to `reconnecting` and feeds a `stream-interrupted` event into the switch tracker. Before this PR the deps were only `active`, `baseUrl`, `targetDevice`, and `useWebRtc`.

Suggested fix: keep the latest `restartWebRtc` in a ref updated with `useLayoutEffect`, the same pattern `deviceScopeRef` uses near line 258, and remove it from these deps.

packages/@expo/hub-client/src/useAndroidDevice.ts
```diff
@@ -849,9 +892,8 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
       setStatus('streaming');
       setError(null);
     } else if (webRtcWasLive && !webRtcGraceExpired) {
-      // When serve-emu swaps the capture source the RTP video usually keeps
-      // flowing; only the control socket is closed and reopened. Keep the frame
-      // and report a reconnect instead of an error while that settles.
+      // A source switch replaces both the control socket and the video peer.
```
krystofwoldrich-agent, line 903: **[suggestion]** The poster is only shown for the 5-second grace window.

The grace timer starts when the peer is torn down. After `STREAM_RECONNECT_GRACE_MS` (5 s) the status falls to `connecting`, or to `error` if the control socket has not reopened. `deviceScreenPresentsMedia` then returns false, and DeviceScreen paints a black surface with a "Connecting…" overlay on top of the poster.

The PR description reports about 9 s total recovery under emulator load. Negotiation of the replacement peer starts after the PATCH returns, so the grace covers only negotiation, and the margin is thin. Consider keeping the media visible while the switch tracker is in `awaiting-frame`, which already has its own 8 s timeout, instead of relying on the grace timer alone.

packages/@expo/hub-client/src/useAndroidDevice.ts
```diff
@@ -814,6 +817,46 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
...
+    // Closing the peer or replacing srcObject can clear the decoded frame.
+    // Capture it before either happens and show it until fresh video arrives.
+    if (video && video.readyState >= 2 && video.videoWidth > 0 && video.videoHeight > 0) {
+      const snapshot = document.createElement('canvas');
+      snapshot.width = video.videoWidth;
+      snapshot.height = video.videoHeight;
+      const context = snapshot.getContext('2d');
+      if (context) {
+        context.drawImage(video, 0, 0);
+        video.poster = snapshot.toDataURL('image/png');
```
krystofwoldrich-agent, line 831: **[suggestion]** `toDataURL('image/png')` is synchronous and encodes the full video resolution on the main thread. The output size grows with the frame size and is stored as a string on the element until the first frame arrives. A JPEG with a quality value, or a smaller canvas, gives the same visual hold at lower cost.

packages/@expo/hub-client/src/useAndroidDevice.ts
```diff
@@ -814,6 +817,46 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
...
+      }
+      video.srcObject = null;
+    }
+    restartWebRtcStream();
+  }, [restartWebRtcStream, webRtcVideoElement]);
+
+  const updateStreamSettings = useCallback(
```
krystofwoldrich-agent, line 844: **[suggestion]** The new tests cover `useWebRtcStream` and `useStreamSettingsResource` directly. The logic here that joins them is untested: the pending guard, the shared `streamSourceRequestRef` token, the success/failure dispatch into the switch tracker, and the poster capture and removal. The socket-close restart path at line 1397 is also untested. A test that drives `useAndroidDeviceClient` with a stubbed `fetch` and `WebSocket` would cover the behavior this PR actually changes.

packages/@expo/hub-client/package.json
```diff
@@ -44,8 +44,10 @@
     "@swc/core": "^1.15.43",
     "@types/bun": "latest",
     "@types/react": "~19.2.2",
+    "@types/react-test-renderer": "19.1.0",
     "oxfmt": "^0.56.0",
     "oxlint": "^1.71.0",
+    "react-test-renderer": "19.2.3",
```
krystofwoldrich-agent, line 50: **[nit]** The test run prints "react-test-renderer is deprecated" three times. This is the first component-render dependency in the package. If the team accepts it, a short note in the test file explaining why it was chosen over a DOM-based runner helps the next person.

krystofwoldrich-agent (changes requested):

**Verdict:** REQUEST_CHANGES
**Findings:** 0 critical · 1 design · 3 suggestion · 1 nit

This is PR 2/2 in the stack on #82.

The change adds a `restart()` to `useWebRtcStream` and calls it from two places in the Android client: when the control socket reports a deliberate server close, and after a stream-settings PATCH succeeds. The settings write now returns a promise so the client can wait for it. The last decoded frame is saved as the video poster so the replacement does not flash black.

I checked serve-emu: a settings change keeps the old peer and calls `resetVideoSource()` on it, while a source switch disposes the session with close code 1012. The two restart triggers match that server behavior. Tests, typecheck, and lint pass on the head commit.

The main concern is the new callback in the control-socket effect's dependency list. It makes that socket reconnect whenever the video element changes. Details are in the inline comments.

**Findings**

- **[design]** `useAndroidDevice.ts:1425` — the control socket reconnects whenever `webRtcVideoElement` changes, including a double connect when WebRTC turns on.
- **[suggestion]** `useAndroidDevice.ts:895` — the poster is only visible during the 5 s grace window; after that DeviceScreen covers it with the Connecting overlay.
- **[suggestion]** `useAndroidDevice.ts:831` — PNG `toDataURL` is synchronous at full resolution; JPEG or a smaller canvas is cheaper.
- **[suggestion]** `useAndroidDevice.ts:838` — the orchestration that joins the two hooks (guard, request token, tracker dispatch, poster) has no test.
- **[nit]** `package.json:50` — `react-test-renderer` prints a deprecation warning in the test run.

packages/@expo/hub-client/src/useAndroidDevice.ts
krystofwoldrich: Fixed in e72b5dd5. The control socket now reads the latest restart callback through a ref updated by useLayoutEffect; video element changes are no longer socket-effect dependencies. The Android-client regression test verifies one socket across initial attachment and remount, then confirms a deliberate close captures the latest video node.

— Codex (GPT-6)

packages/@expo/hub-client/src/useAndroidDevice.ts
krystofwoldrich: Fixed in e72b5dd5. An already-live stream stays Reconnecting while the switch tracker awaits a frame, even after the ordinary five-second grace expires. The existing eight-second frame timeout remains the limit. The Android-client test verifies media stays visible past grace, the timeout releases pending state, and a fresh frame clears the poster.

— Codex (GPT-6)

packages/@expo/hub-client/src/useAndroidDevice.ts
krystofwoldrich: Added Android-client integration tests in e72b5dd5 using the real hooks with controlled fetch, WebSocket, peer, video, and timer events. They cover initial-loading and overlapping-write guards, stale device responses, successful/failed writes, source commits after fresh frames, deliberate socket-close restarts, remounts, poster capture/removal, and bounded frame waits. Both reported behavior bugs failed these tests before the fix. All 210 client/component tests pass.

— Codex (GPT-6)

packages/@expo/hub-client/package.json
krystofwoldrich: Documented the choice in the test file in e72b5dd5: the renderer exercises hook lifecycles with controlled transport/media events in the existing Bun runner without adding a DOM environment. The deprecation warning remains visible; real decoding and presentation are covered by the browser recording.

— Codex (GPT-6)

packages/@expo/hub-client/src/useAndroidDevice.ts
krystofwoldrich: Deferred to a follow-up as requested. PNG snapshot conversion is unchanged in this PR; the follow-up is also noted in the description.

— Codex (GPT-6)

PR #87, opened 2026-09-09
feat(hub): add an Accessibility inspector section for iOS and Android

vonovak:
## Why

The Hub has no accessibility section for either platform. Both backends already expose an accessibility tree: serve-emu through `GET /api/accessibility` (a `uiautomator dump`) and serve-sim through its `/ax` SSE route. One Hub section serves both, so an agent or a person can read the labels on the current screen and tap an element by name instead of by pixel.

## Scope

- `@expo/hub-client`
  - `types.ts` adds `AccessibilityNode`, `AccessibilitySnapshot`, `AccessibilityFrame`, `DeviceCapabilities.accessibility`, and four `DeviceClient` members: `accessibility`, `accessibilityPending`, `accessibilityError`, `refreshAccessibility()`. Frames are 0..1 screen fractions, so a tap goes through the existing `sendTouch`.
  - `accessibility.ts` holds the two pure parsers, `parseAndroidAccessibility` and `parseIosAccessibility`, and the two loaders. `normalizedFrame` is the one place that converts a backend rectangle to screen fractions: it clamps the origin into the 0..1 box and caps the size at the space left, because serve-sim reports frames that hang off the display and `uiautomator` reports negative bounds for scrolled rows. The Android parser divides by the widest node, because the `uiautomator` root spans the display and the payload carries no screen size. The iOS parser divides by the reported `screen`. Both drop elements with no name.
  - `sse.ts` receives `drainSseChunk` from `useIosDevice.ts` unchanged and adds `readSseSnapshot`, a one-shot read of an SSE route with a settle window. serve-sim replays its cached snapshot on connect and writes the fresh poll result after it, so the reader holds the stream open for `IOS_AX_SETTLE_MS` after the first block and returns the newest. A mid-stream failure rethrows instead of passing off that first block as a fresh snapshot.
  - `useAccessibility.ts` owns the state for both platforms. Each platform hook builds an `AccessibilityLoader` or `null`, and `capabilities.accessibility` is derived from that same value, so the flag and the action cannot disagree.
  - `useAndroidDevice.ts`, `useIosDevice.ts` and `useNoopDeviceClient.ts` gain the wiring. The iOS hook reads the `axEndpoint` field the serve-sim middleware `/api` config already returns.
- `@expo/hub-components`
  - `AccessibilitySection.tsx`: a Refresh row, one status note, and a scrolling list of elements with label and role. A click sends `begin` then `end` at the frame center. Registered in `LogSidebar.tsx` after Camera, gated on `capabilities.accessibility`.
- `expo-device-hub`: `inspectorSections.test.tsx` pins the new section's position on both platforms. One `minor` changeset.

Out of scope: a highlight overlay on the device screen, a backend tap route, and turning accessibility on inside the simulator.

## Tradeoffs

- State is the neighbor triple that `camera`, `cameraPending` and `cameraError` already use, not a discriminated union. Every combination is meaningful, because a stale list stays on screen while a refresh runs or after one fails.
- The section refreshes on open and on demand only. serve-emu's read runs `uiautomator dump` for up to 8 s on the interactive adb lane, so a poll would starve input.
- `IOS_AX_SETTLE_MS` is 2000 ms. Measured against the booted simulator, the fresh `/ax` block lands 0.68 s to 1.36 s after the cached one, and a 1000 ms window returned the stale snapshot on the slower connects. The window is still a guess at a protocol: serve-sim marks neither block, so when the native poll runs longer than 2 s the Hub shows the cached tree and stamps it with the current time. Fixing that needs a `capturedAt` or a replay marker from serve-sim.
- iOS `clickable` comes from a role allowlist, `IOS_CLICKABLE_ROLES`, because serve-sim reports no actionable flag. The list was checked against the roles the Settings app exposes. Android `clickable` is the real attribute, so the two platforms answer a slightly different question and the "tappable" tag is advisory on iOS.
- The iOS read is a one-shot HTTP read of the SSE route, not a standing exec-ws subscription. The helper's raw `GET /helper/<udid>/ax` was rejected because it returns the un-normalized native tree.
- The Android list is noisy. When a node has no `text` and no `contentDescription`, the parser falls back to the `resourceId` tail, so layout containers such as `icon_frame` and `widget_frame` appear beside real rows. Keeping them preserves the only handle those nodes have, and some carry the `clickable` flag for the row they wrap. Filtering to clickable-or-named is the obvious follow-up and wants its own live comparison.

## Blast Radius

The change is additive on `DeviceClient`, so every implementer and test literal gained the new members in this PR. `drainSseChunk` moved between files with no change to its body; its only caller is the iOS exec-ws multiplexer, and `sse.test.ts` covers it. No serve-emu or serve-sim change, so no backend PR. The `.changeset` entry names only `expo-device-hub`.

An adversarial review pass raised four issues this PR does not close. They are worth a reviewer's attention and each needs its own evidence:

- Android normalization divides by the widest node. `server.ts` floors the same computation with the real display size before it taps. When a dump does not span the display, every coordinate scales up and a tap lands low and right. The fix is to pass the `screen` the Android hook already holds into the loader.
- Neither loader carries a client-side deadline, so Refresh stays disabled for as long as the backend takes. Android is bounded by the server at roughly 25 s; iOS is not bounded at all.
- Aborting the fetch does not cancel the server-side dump. The route calls `getAccessibility` without the request signal, so a device switch queues a second dump behind the first on the shared interactive adb lane. That is a serve-emu contract change.
- A node whose center sits in the bottom 7% of an iOS screen is routed to the home-indicator recognizer by `homeIndicatorEdge`, so its tap reads as a swipe to home. Clamping bounds the frame but does not solve this.

## Verification

Gates from the worktree root, all green: `bun run lint` (4/4 packages, 0 warnings), `bun run typecheck` (11/11), and `bun run test --force` (575 pass, 0 fail across `@expo/hub-client` 207, `expo-device-hub` 174, `@expo/hub-android-utils` 121, `@expo/hub-apple-utils` 44 and `@expo/hub-components` 29). `bunx tsc --noEmit` in `packages/expo-device-hub` reports zero errors outside `vendor/`. `bunx changeset status` bumps only `expo-device-hub`.

Unit coverage: 20 tests over both parsers and both loaders in `accessibility.test.ts`, 10 over `drainSseChunk` and `readSseSnapshot` in `sse.test.ts`, and 5 over the section in `accessibility-section.test.tsx`. The loaders take an injected `fetchImpl`, so no global is patched. Two of those tests were written against a bug found in review and fail without its fix: the mid-stream SSE failure, and the message serve-emu actually sends, which is `{ok: false, error: {code, message}}` and used to render as "Malformed accessibility response".

End to end in the Hub website, driven with Playwright against real devices: the shared `emulator-5554` and a booted iPhone 17 Pro simulator. The script opens the Hub, picks the device, expands Accessibility, reads the rows, taps one by its `aria-label`, and proves the device moved by a readback taken outside the browser.

- Android: 58 rows off the Settings home screen. Tapping the `Network & internet` row moved `adb shell dumpsys window | grep mCurrentFocus` from `com.android.settings.homepage.SettingsHomepageActivity` to `com.android.settings.SubSettings`, the `adb exec-out screencap` bytes changed, and Refresh returned a different 43-row list in 3.7 s, led by a `Network & internet` title and a `Navigate up` button.
- iOS: 20 rows off the Settings root. Tapping `General` changed the `simctl io screenshot` bytes, and Refresh returned a different 17-row list in 1.3 s, led by `Settings | back button`, which is the post-navigation screen and not the cached snapshot.

The simulator needs `AccessibilityEnabled` and `ApplicationAccessibilityEnabled` set, or the native bridge reports an empty 0x0 hierarchy. The e2e runner sets both and restores the prior values on exit. The runner does not ship in this PR.


vonovak:

- Gates rerun in a fresh worktree at `beda760`: `bun run lint`, `bun run typecheck`, `bun run test --force` all exit 0; `tsc --noEmit` in `packages/expo-device-hub` shows no error outside `vendor/`; `changeset status` bumps `expo-device-hub` only.
- Receipts: 16 files, new `accessibility.ts`, `sse.ts`, `useAccessibility.ts` and `AccessibilitySection.tsx`; shared-file edits are additive; taps go through `sendTouch` at the frame centre; tokens only.
- Runtime, independent rerun from the verifier worktree on port 8099: Android read 58 rows on Settings home, tapping `Network & internet` moved `mCurrentFocus` from `SettingsHomepageActivity` to `SubSettings`, the screencap changed, and Refresh returned 43 rows in 3.6 s. iOS read 20 rows, tapping `General` changed the simulator screenshot and Refresh returned 17 rows in 1.3 s.
- Rebased onto `vonovak/android-multitouch` (link 1) with an identical patch-id, so the verdict carries over.

packages/@expo/hub-client/src/accessibility.ts
```diff
@@ -0,0 +1,219 @@
...
+  bottom: number,
+  extentX: number,
+  extentY: number,
+): AccessibilityFrame {
+  const x = Math.min(Math.max(left / extentX, 0), 1);
+  const y = Math.min(Math.max(top / extentY, 0), 1);
+  return {
+    x,
+    y,
+    width: Math.min((right - left) / extentX, 1 - x),
```
krystofwoldrich-agent: **[design]** `x` is clamped into 0..1, but `width` is `min((right - left) / extentX, 1 - x)`, the full width of the element. When an element starts above or left of the screen, the frame keeps its full size from the clamped origin, and the centre that `tapNode` computes moves off the visible part.

Example on an 844 pt iOS screen: `frame.y = -30, height = 44` becomes `y = 0, height = 44/844`, centre 22 pt, while the visible span is 0..14 pt. serve-sim `normalizeAxTree` (`ax.ts`) passes element frames through unchanged, so partly scrolled cells arrive in this shape. An element fully above the screen ends with zero height and still renders as a tappable row.

Fix: clamp both edges (`x0 = clamp(left / extentX)`, `x1 = clamp(right / extentX)`, `width = x1 - x0`), same for `y`, and drop nodes whose clamped area is zero. The iOS clamp test keeps passing with two-edge clamping; the Android test at `accessibility.test.ts:145` encodes the current result and needs the same update.

Note: serve-emu `parseBounds` matches `\[(\d+),(\d+)\]\[(\d+),(\d+)\]`, non-negative digits only, so today the Android path cannot produce negative bounds. The iOS path can.

packages/@expo/hub-client/src/__tests__/accessibility.test.ts
```diff
@@ -0,0 +1,317 @@
...
+  test('clamps a node scrolled off the top left', () => {
+    const nodes = snapshotOf(
+      parseAndroidAccessibility(
+        androidBody([
+          ANDROID_ROOT,
+          androidNode({ text: 'Scrolled out', bounds: { left: -40, top: -20, right: 200, bottom: 100 } }),
+        ]),
+      ),
+    ).nodes;
+    expect(nodes[0].frame).toEqual({ x: 0, y: 0, width: 240 / 1080, height: 120 / 2400 });
```
krystofwoldrich-agent: **[design]** This expectation encodes the one-sided clamp. The visible extent of `left: -40, right: 200` is 200 px, so with both edges clamped the width is `200 / 1080` and the height `100 / 2400`. See the comment on `normalizedFrame`.

packages/@expo/hub-components/src/dashboard/AccessibilitySection.tsx
```diff
@@ -0,0 +1,120 @@
...
+  if (snapshot.nodes.length === 0) {
+    return <SectionNote>No accessible elements on this screen.</SectionNote>;
+  }
+  return <SectionNote>{`Captured ${new Date(snapshot.capturedAt).toLocaleTimeString()}`}</SectionNote>;
+}
+
+function tapNode(client: DeviceClient, node: AccessibilityNode) {
+  const x = node.frame.x + node.frame.width / 2;
+  const y = node.frame.y + node.frame.height / 2;
+  client.sendTouch({ phase: 'begin', x, y });
```
krystofwoldrich-agent: **[design]** `useIosDeviceClient.sendTouch` (`useIosDevice.ts:454-456`) tags every `begin` with `y >= HOME_INDICATOR_BAND_NORM` (0.93, `orientation.ts:26`) with `HID_EDGE_BOTTOM`. This `begin` lands at the frame centre, so any element whose centre is in the bottom 7% of the screen (bottom-anchored buttons, tab bars) is sent as a home-indicator edge gesture, not a plain tap. The PR body reports that the device reads this as a swipe to home. The section's one action is wrong for a common class of targets.

Options: when the frame extends above the band, pick a point inside the frame above `HOME_INDICATOR_BAND_NORM`; or add a way for a programmatic tap to opt out of edge detection in `sendTouch`.

packages/@expo/hub-client/src/sse.ts
```diff
@@ -0,0 +1,79 @@
...
+  signal.addEventListener('abort', abort);
+  let settleTimer: ReturnType<typeof setTimeout> | null = null;
+
+  try {
+    const response = await fetchImpl(url, {
+      cache: 'no-store',
+      headers: { accept: 'text/event-stream' },
+      signal: controller.signal,
+    });
+    if (!response.body) return null;
```
krystofwoldrich-agent, line 54: **[suggestion]** serve-sim answers `/ax` with `404` and the plain-text body `No serve-sim device` when the device is not selected. This reader drains that body as SSE, finds no block, returns `null`, and the section shows "No accessibility snapshot received". `loadAndroidAccessibility` reports the HTTP status for the same case. Throw or return an `HTTP <status>` error when `!response.ok`.

packages/@expo/hub-client/src/accessibility.ts
```diff
@@ -0,0 +1,219 @@
+import { type SseFetch, readSseSnapshot } from './sse';
+import {
+  type AccessibilityFrame,
+  type AccessibilityNode,
+  type AccessibilitySnapshot,
+} from './types';
+
+/** serve-sim replays a cached snapshot on connect and writes the fresh poll ~1.3 s later. */
+export const IOS_AX_SETTLE_MS = 2000;
```
krystofwoldrich-agent, line 12: **[suggestion]** `createAxStreamer.poll` in serve-sim `ax.ts` writes the fresh snapshot only when `nextMessage !== latestMessage`. On an unchanged screen the second block never arrives, so every iOS Refresh takes the full 2 s and the Refresh button stays disabled for that time. This comment holds only when the tree changed; state that here. The real fix is a replay marker or `capturedAt` from serve-sim, out of scope as the PR body says.

packages/@expo/hub-client/src/useAccessibility.ts
```diff
@@ -0,0 +1,55 @@
...
+  const refreshAccessibility = useCallback(() => {
+    controllerRef.current?.abort();
+    controllerRef.current = null;
+    if (!load) return;
+
+    const controller = new AbortController();
+    controllerRef.current = controller;
+    setAccessibilityPending(true);
+    setAccessibilityError(null);
+    load(controller.signal)
```
krystofwoldrich-agent: **[suggestion]** The loaders run until the backend answers. serve-emu `dumpXml` tries `DUMP_ATTEMPTS = 3` times, each with two adb calls at `timeout: 8_000` plus a backoff, so one stuck emulator can hold `accessibilityPending` for about 48 s with Refresh disabled. `/ax` has no server bound. Wrap the signal, for example `AbortSignal.any([signal, AbortSignal.timeout(ms)])`, and show a timeout message.

packages/@expo/hub-client/src/accessibility.ts
```diff
@@ -0,0 +1,219 @@
...
+  if (body.ok === false) return { ok: false, error: failureText(body.error) || MALFORMED };
+
+  const capturedAt = Date.parse(str(body.capturedAt));
+  if (body.ok !== true || !Array.isArray(body.nodes) || !Number.isFinite(capturedAt)) {
+    return { ok: false, error: MALFORMED };
+  }
+
+  const entries: { raw: Record<string, unknown>; bounds: AndroidBounds }[] = [];
+  // The snapshot carries no screen size, but uiautomator's root node spans the
+  // display, so the widest bounds are the extent to normalize against.
```
krystofwoldrich-agent: **[suggestion]** serve-emu's own tap route (`server.ts:1163-1172`) floors the same computation with `context.screen.width/height`. Here only `max(bounds.right)` and `max(bounds.bottom)` are used. When a dump does not span the display, fractions inflate and taps drift right and down. The PR body proposes passing the hook's `screen`. Note that on Android `screen` is set from decoded frame sizes (`useAndroidDevice.ts:1130`, `:1193`, `:1350`), so confirm it equals device pixels before using it as the floor.

packages/@expo/hub-client/src/useAccessibility.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import { type AccessibilityLoader } from './accessibility';
+import { type AccessibilitySnapshot, type DeviceClient } from './types';
+
+type AccessibilityClientState = Pick<
+  DeviceClient,
+  'accessibility' | 'accessibilityPending' | 'accessibilityError' | 'refreshAccessibility'
+>;
+
+export function useAccessibility(load: AccessibilityLoader | null): AccessibilityClientState {
```
krystofwoldrich-agent: **[suggestion]** This hook holds the most delicate logic in the PR: abort on loader change, the cleanup-order comment, and the `controllerRef === controller` guard in `finally`. `webrtc-stream-restart.test.tsx` already uses `react-test-renderer` in this package for hook lifecycles. A test for a loader change during a read, and for the pending reset, would pin the comment's claim. `tapNode`'s centre math is also untested; the section tests render with `renderToStaticMarkup`, which cannot click.

packages/@expo/hub-client/src/accessibility.ts
```diff
@@ -0,0 +1,219 @@
...
+      frame: normalizedFrame(x, y, x + frameWidth, y + frameHeight, width, height),
+    });
+  }
+  return { ok: true, snapshot: { capturedAt, nodes } };
+}
+
+export async function loadAndroidAccessibility(
+  url: string,
+  signal: AbortSignal,
+  fetchImpl: SseFetch = fetch,
```
krystofwoldrich-agent: **[nit]** `loadAndroidAccessibility` takes `fetchImpl: SseFetch` but reads JSON. A neutral name such as `FetchLike` in `sse.ts` would read better for both callers.

krystofwoldrich-agent (changes requested):

**Verdict:** REQUEST_CHANGES
**Findings:** 0 critical · 2 design · 5 suggestion · 1 nit

The PR adds an Accessibility section to the sidebar. Two pure parsers turn the serve-emu `GET /api/accessibility` body and the serve-sim `/ax` SSE payload into one node model in 0..1 screen fractions, and a shared `useAccessibility` hook owns the state. The structure follows the existing camera pattern and the capability flag is derived from the loader. The main concern is the tap path: `normalizedFrame` clamps only the origin, so a partly off-screen element gets a centre outside its visible area, and `tapNode` sends its `begin` through the iOS home-indicator band check, so bottom-anchored elements are sent as edge gestures. Both are fixable inside this PR.

Checks run in a worktree: `tsc --noEmit` passes for `@expo/hub-client` and `@expo/hub-components`. `bun test` passes for `@expo/hub-client` (224), `@expo/hub-components` (29), and `expo-device-hub` (177) after the workspace packages were built.

### Findings

- **[design]** `normalizedFrame` clamps the origin but not the far edge, so a partly off-screen element taps outside its visible part — `packages/@expo/hub-client/src/accessibility.ts:67`
- **[design]** `tapNode` taps whose centre is in the bottom 7% are sent as home-indicator edge gestures — `packages/@expo/hub-components/src/dashboard/AccessibilitySection.tsx:30`
- **[suggestion]** `readSseSnapshot` does not check `response.ok`; a 404 shows as "No accessibility snapshot received" — `packages/@expo/hub-client/src/sse.ts:53`
- **[suggestion]** The settle window waits for a second block that serve-sim sends only when the tree changed; unchanged screens take the full 2 s — `packages/@expo/hub-client/src/accessibility.ts:9`
- **[suggestion]** No client-side deadline on a read; serve-emu can hold the pending state for about 48 s — `packages/@expo/hub-client/src/useAccessibility.ts:26`
- **[suggestion]** Android extent comes from the widest node only; serve-emu's tap route floors it with the real screen — `packages/@expo/hub-client/src/accessibility.ts:112`
- **[suggestion]** No lifecycle test for `useAccessibility` although the package already uses `react-test-renderer` for hooks — `packages/@expo/hub-client/src/useAccessibility.ts:11`
- **[nit]** `SseFetch` names the JSON fetch in `loadAndroidAccessibility` — `packages/@expo/hub-client/src/accessibility.ts:191`

vonovak:

**Act on**
- Android frames are normalized against `max(bounds.right)` and `max(bounds.bottom)` of the dump, not the screen. Off-screen or clipped nodes inflate the divisor and every tap drifts, silently. Pass the hook's screen size into `parseAndroidAccessibility` and keep the node maximum as a fallback; confirm the Android `screen` value is device pixels, since it comes from decoded frame sizes. (interrogate reviewer A and the automated review, independently)
- One-sided clamp in `normalizedFrame`: the origin is clamped into 0..1 but the size keeps the element's full extent, so a partly scrolled iOS cell gets a centre outside its visible part, and a fully off-screen element still renders as a tappable row. Clamp both edges and drop zero-area nodes; update the Android clamp test at `accessibility.test.ts:145`. (automated review)
- iOS taps in the bottom 7 percent of the screen become a home-indicator edge gesture, because `useIosDeviceClient.sendTouch` tags any `begin` with `y >= 0.93` as `HID_EDGE_BOTTOM`. Tab bars and bottom buttons are the common case. Pick a point inside the frame above the band, or let a programmatic tap opt out of edge detection. (automated review)
- serve-sim answers `/ax` for an unselected device with a 404 text body; `readSseSnapshot` drains it as SSE and the section shows a generic "no snapshot" note. Report the HTTP status like the Android loader does. (automated review)

**Consider**
- The loaders have no time bound; one stuck emulator can hold `accessibilityPending` for about 48 seconds. Wrap the signal with a timeout and show a message. (automated review)
- `useAccessibility` and `tapNode` have no test. `webrtc-stream-restart.test.tsx` already uses `react-test-renderer` in this package, so a loader-change-during-read test is feasible. (automated review; interrogate B flagged the untested hooks stack-wide)

**Noted**
- The 2 s settle only matters when the tree changed; serve-sim sends the fresh block only on change. Say so in the comment. (automated review)
- `SseFetch` as the type name for a JSON loader reads oddly. (automated review nit)

vonovak:
Fix round for the combined review triage. New head `ab9c25b`, five commits on top of `8b63138`.

**Act on, fixed**
- Two-edge clamp, zero-area nodes dropped: `07338e7`. `visibleFrame` clamps both edges into 0..1 and returns null when nothing is visible; both parsers skip null frames. The Android clamp test at `accessibility.test.ts` now expects `200/1080` by `100/2400`.
- Android screen-size normalization: `0e491ea` (serve-emu) and `ab9c25b` (hub-client). The confirmation the item asked for came back negative: `adb shell wm size` is 1080x2400, while every writer of the hook's `screen` on Android is the stream size (the paint canvas measured 576x1280; the `/api` poll writes serve-emu `context.screen`, which serve-emu sets from the scrcpy session frame). So the hook's `screen` cannot be the floor. Instead the serve-emu snapshot now reports `screen` from the effective `wm size` (override preferred, the space uiautomator reports in), swapped by the dump's own `rotation` attribute, and falls back to the dump extent when the size query is refused. `parseAndroidAccessibility` requires `screen` and normalizes against it, as the iOS parser already does. The e2e reads the proxied `/api/accessibility` and asserts `screen` equals `wm size`.
- iOS bottom-band taps: `e5494d9`. `TouchSample.edgeGestures?: boolean`; `homeIndicatorEdge` returns undefined when it is false; `tapNode` passes false on `begin`. Chosen over picking a point above the band because a row fully inside the band has no such point. Pointer input from `DeviceScreen` is unchanged. E2E: the Settings search field (centre y 0.9405, inside the 0.93 band) tapped from the section; the serve-sim "Current app" readback stayed `com.apple.Preferences` and the refreshed tree shows the search UI with a keyboard.
- HTTP status on `/ax`: `163d359`. `readSseSnapshot` throws `HTTP <status>` on `!response.ok`; the section shows it through the hook's catch.

**Left**
- Consider and Noted items (loader time bound, hook tests, settle comment, `SseFetch` name): out of this round's scope.
- The single interrogate reviewer also flagged that serve-emu's own UI panel (`src/ui/components/device-stream.tsx`) still measures hover hit-testing against the video size. Pre-existing and outside the act-on fence; `snapshot.screen` is now available for it.

**Gates on `ab9c25b`**: Hub `bun run lint`, `bun run typecheck`, `bun run test --force` green; Hub `tsc` shows only the three pre-existing `serve-emu-options.test.ts` errors from main. serve-emu `typecheck`, `bun test` (934), `docs:check` green; `build:vendor` rerun and the vendored `accessibility.js` carries `snapshotFromDump`. E2E run 10 under the lane: PASS on both devices (Android focus `SettingsHomepageActivity` to `SubSettings`, 58 to 43 rows; iOS band tap plus `General` tap, 20 to 17 rows).

vonovak:
## Fix round and restack

All four act-on items landed at `436dfd3`: both edges of a frame are clamped and empty nodes dropped, a non-2xx `/ax` reports its HTTP status, a targeted tap opts out of the iOS home-indicator edge gesture, and the Android extent now comes from serve-emu's reported display size. The owner disproved the scope's premise first: the Hub's Android `screen` is the paint canvas (576x1280), not device pixels (1080x2400), so the fix moved to the backend, which reads the effective `wm size` and swaps axes by the dump's rotation. The e2e gained an iOS band-tap case.

krystofwoldrich-agent (approved):
Question: Was rendering an accessibility overlay over the device frame, such as bounding boxes or highlights for the listed elements, out of scope for this PR?

vonovak:
> Question: Was rendering an accessibility overlay over the device frame, such as bounding boxes or highlights for the listed elements, out of scope for this PR?

out of scope, not even ios does it now in the hub

PR #88, opened 2026-09-09
feat(hub): add an Android on-screen keyboard control

vonovak:
## Why

`ANDROID-PARITY.md` item 2 asks for the Android half of the keyboard controls. Android
exposes one secure setting, `show_ime_with_hard_keyboard`. Writing `1` shows the on-screen
IME even while a hardware keyboard is attached; writing `0` restores the default, where the
IME hides. That is persistent, readable, pollable boolean state, which is exactly what the
`DeviceSettingKey` union and the `ANDROID_DEVICE_SETTINGS` registry already model seven
times over.

So the Android control enters the registry as a device setting, not the `DeviceClient`
keyboard trio. `setDeviceSetting`, the optimistic write, the pending flag, the 3 s poll and
the stale-scope guard all come for free. `DeviceClient` gains no member, and
`useAndroidDevice.ts`, `useNoopDeviceClient.ts` and `useIosDevice.ts` are untouched.

## Scope

serve-emu gains `GET/POST /api/software-keyboard`, modelled on `/api/font-weight`.

- `src/adb.ts`: `getSoftwareKeyboard` and `setSoftwareKeyboard` read and write
  `secure show_ime_with_hard_keyboard` through the existing `secureSetting`, `mutateShell`
  and `enabledFromIntSetting` helpers. The setter returns a readback, not an echo.
- `src/shared/api-contracts.ts`: `SoftwareKeyboardStatus`, `SoftwareKeyboardResponse`, the
  `ApiContractMap` pair, `parseSoftwareKeyboardResponse`, and the `API_SUCCESS_PARSERS`
  entry.
- `src/middleware.ts` and `src/server.ts` serve the pair; the middleware POST keeps the
  `isAllowedBrowserOrigin` check its siblings use.
- `src/api/routes/devices.ts` and `src/api/dependencies.ts` extend the typed route table.
  It now holds 58 method/path pairs across 40 paths.
- `README.md` documents the curl pair and what the key does; `docs:sync` regenerates the
  package README.

The Hub adds one setting key, `onscreen-keyboard`.

- `hub-client/src/types.ts`: one member added to `DeviceSettingKey`.
- `hub-client/src/android-device-settings.ts`: the `AndroidDeviceSettingKey` and
  `AndroidDeviceSettingPath` members plus the registry entry mapping
  `onscreen-keyboard` to `/api/software-keyboard` with `polled: true`. Both records are
  total, so a half-registered key fails the build.
- `hub-components/src/dashboard/DeviceOptionsSection.tsx`: the `DEFAULT_VALUES` entry, the
  exported `ONSCREEN_KEYBOARD_DESCRIPTION`, and one `SidebarRow` block gated on
  `showDeviceSettings && platform === 'android' && visible('onscreen-keyboard')`.

Out of scope: the iOS trio, `hardwareKeyboardConnected` plus its two setters. Replacing it
with one grouped nullable member is the right end state, but it costs nine deletions in
three hot zones that six concurrent branches are appending to, and it improves code no user
touches. It is a clean feature-free PR on its own after the restack.

## Tradeoffs

The row stays out of `SWITCH_OPTIONS`. `visible(key)` returns true for every key on iOS
while `deviceSettings` is still `null`, which is a live window during every iOS connect
(`useIosDevice.ts` sets `capabilities.deviceSettings` before the first payload lands). A
`SWITCH_OPTIONS` entry would flash an Android-only switch on the iOS sidebar. The
standalone block puts `platform === 'android'` before `visible`, so that branch is
unreachable by construction. It copies the display-size row 15 lines above. The cost is
about eight duplicated lines of switch JSX.

The label is "On-screen keyboard", not "Software keyboard". A switch called "Software
keyboard" claims off means no software keyboard. Off restores Android's default. The
description carries that qualifier and is wired to `aria-describedby`. The iOS button keeps
its own label; the two controls have different affordances and different readback.

## Blast Radius

The registry, `DEFAULT_VALUES` and the setting union are total records, so the compiler
rejects a partially registered key. No existing row moves: the new block is gated to
Android, and every order-pinning assertion in `inspectorSections.test.tsx` renders an iOS
client. `DeviceClient` is unchanged, so both `DeviceClient` fixtures still compile
untouched.

For a reviewer restacking seven branches: this PR deletes no line in any shared hot zone.
The only shared-file edit is one added line in the `DeviceSettingKey` union.

## Verification

serve-emu gates in `packages/serve-emu/packages/serve-emu`: `bun run typecheck` clean (it
covers `tsconfig.tests.json` too), `bun test` 917 pass 0 fail across 100 files,
`bun run docs:check` reports the package README synchronized.

Hub gates from the repo root: `bun run lint` 0 warnings 0 errors, `bun run typecheck` 12
tasks successful, `bun run test --force` green in every package (hub-client 177,
hub-components 24, expo-device-hub 176, hub-android-utils 121, hub-apple-utils 44).
`cd packages/expo-device-hub && bunx tsc --noEmit` prints 0 non-vendor errors and the 206
known `vendor/` errors. `bunx changeset status` bumps only `expo-device-hub`, at minor.

After `build:vendor`, `vendor/serve-emu/dist/middleware.js` carries the literal
`softwareKeyboard` and `/api/software-keyboard`, and `vendor/serve-emu/dist/adb.js` carries
`show_ime_with_hard_keyboard`. A misspelled envelope key reads back as `null` in the Hub
with no error, so this check is explicit.

End to end in the Hub website, driven with Playwright against `emulator-5554`:

1. Seed the device with `adb shell settings put secure show_ime_with_hard_keyboard 1`, then
   load the Hub. The switch renders checked, so the poll reads the device state through the
   route.
2. Toggle off in the Hub. `adb shell settings get secure show_ime_with_hard_keyboard`
   returns `0` within 776 ms.
3. Toggle on in the Hub. The same read returns `1` within 572 ms.
4. The description renders in the sidebar. The setting is restored to `0` afterwards.

**Known verification gap.** The shared AVD `serve_emu_cam` has `hw.keyboard=no`. Android
only consults `show_ime_with_hard_keyboard` when a hardware keyboard is attached, so on that
AVD the setting cannot change what is on screen. A probe at both values left
`dumpsys input_method` reporting `mInputShown=true` with the IME visible. The proof above is
therefore the settings readback, not a screenshot pair. Visual proof needs an AVD built with
`hw.keyboard=yes`.


vonovak:

- Gates rerun in a fresh worktree at `00e69c7`: Hub `bun run lint`, `bun run typecheck`, `bun run test --force` exit 0; `tsc --noEmit` in `packages/expo-device-hub` shows no error outside `vendor/`; serve-emu `bun run typecheck` 0, `bun test` 917 pass 0 fail, `docs:check` 0; the built `vendor/serve-emu/dist/middleware.js` carries `softwareKeyboard`; `changeset status` bumps `expo-device-hub` only.
- Receipts: the route mirrors `/api/font-weight` across `adb.ts`, `middleware.ts`, `server.ts`, the typed table and the contracts; the Hub side is one `ANDROID_DEVICE_SETTINGS` row plus one Android-gated switch row. No `DeviceClient` change.
- Runtime, independent rerun from the verifier worktree on port 8100 against `emulator-5554`: the switch renders on after seeding the secure setting to `1`; toggling off in the Hub reads `0` back in 580 ms; toggling on reads `1` back in 593 ms; the setting was restored to `0`. As the PR says, this AVD has `hw.keyboard=no`, so the visual half cannot be shown there.
- Rebased onto `vonovak/accessibility-tree` (link 2); the verdict carries over.

vonovak:
Restack note: rebasing onto link 2 conflicted only in `inspectorSections.test.tsx` (both links append tests at the end of the file). The root kept both blocks and added `accessibility: false` to the new capability literal. Hub lint, typecheck, and the full test run pass on the rebased head. The three `tsc --noEmit` errors in `serve-emu-options.test.ts` also exist on `main` since #80 and are not from this stack.

krystofwoldrich-agent (commented):

**Verdict:** COMMENT
**Findings:** 0 critical · 0 design · 1 suggestion · 2 nit

The PR adds `GET/POST /api/software-keyboard` to serve-emu (adb read/write of `secure show_ime_with_hard_keyboard`, contract, parser, typed route, middleware and CLI server blocks, docs) and one Hub setting key `onscreen-keyboard` (union member, Android registry entry, `DEFAULT_VALUES`, one platform-gated `SidebarRow`). Every server piece copies the `/api/font-weight` sibling one-for-one and the Hub piece rides on the existing `ANDROID_DEVICE_SETTINGS` table, so the change is low risk. The main concern is small: the new switch row is a hand-copied block rather than a table entry. Observation, not a finding: this is the seventh `polled: true` key, so each Hub client now issues seven `adb shell settings get` calls every 3 s; that is the existing pattern, but the cost grows with each setting the stack adds.

Checks run at the PR head: serve-emu `typecheck` clean, `docs:check` synchronized, the five touched serve-emu test files 90 pass / 0 fail; hub-client `tsc` clean, 224 tests pass; hub-components `tsc` clean, 29 tests pass; `expo-device-hub` `inspectorSections.test.tsx` 46 pass.

**Findings**
- [suggestion] `packages/@expo/hub-components/src/dashboard/DeviceOptionsSection.tsx:218` — carry the platform gate (and description) in `SWITCH_OPTIONS` instead of copying the switch JSX; the `visible()` null-window problem applies to every future Android-only switch.
- [nit] `packages/@expo/hub-components/src/dashboard/DeviceOptionsSection.tsx:96` — docblock still says "plus iOS keyboard controls"; the section now also renders an Android keyboard setting.
- [nit] PR metadata: commit `f5366db` says every emulator reports a hardware keyboard, while the PR body reports `hw.keyboard=no` on the tested AVD with no visible effect. Both cannot hold; the commit message will mislead `git log` readers. The body also cites `ANDROID-PARITY.md`, which is not present at the PR head.

packages/@expo/hub-components/src/dashboard/DeviceOptionsSection.tsx
```diff
@@ -210,6 +215,21 @@ export function DeviceOptionsSection({
           ) : null,
         )}
 
+      {showDeviceSettings && platform === 'android' && visible('onscreen-keyboard') && (
```
krystofwoldrich-agent, line 222: **[suggestion]** The PR body explains why this is not a `SWITCH_OPTIONS` entry: `visible()` returns true for every key while an iOS client has `deviceSettings === null`, so a table entry would flash an Android-only switch on iOS. That is correct, but the same problem applies to every future Android-only switch. Adding optional `platform?: DevicePlatform` and `description?: string` fields to the `SWITCH_OPTIONS` entries and checking `(!option.platform || option.platform === platform) && visible(key)` inside the existing `.map` fixes it once, keeps one switch renderer, and removes the ~14 duplicated lines. Not blocking.

vonovak:

**Consider**
- The new switch row duplicates the `SWITCH_OPTIONS` renderer to avoid the iOS flash while `deviceSettings` is null. An optional `platform` (and `description`) field on the table entries, checked inside the existing map, fixes it once for every future Android-only switch and removes the duplicated block. (automated review; all three interrogate reviewers noted the duplication)

**Noted**
- The iOS keyboard trio (`hardwareKeyboardConnected`, `setHardwareKeyboardConnected`, `toggleSoftwareKeyboard`) stays as it was; the parity backlog asked for a rethink and the PR body defers it. (interrogate; disclosed)
- The shared AVD has `hw.keyboard=no`, so verification rests on the settings readback, not a visible keyboard. (disclosed in the PR body)

No correctness findings from either source on this link.

vonovak:
## Fix round and restack

No act-on items for this link. The Consider item (a `platform` field on the `SWITCH_OPTIONS` entries so an Android-only switch needs no duplicated renderer) stays a follow-up. Head is now `adba8f3`.

krystofwoldrich-agent (approved):
[nit] When hw.keyboard=no, we should disable the on-screen keyboard toggle since changing this setting has no visible effect without a hardware keyboard. A short explanation alongside the disabled control would make that clear.

vonovak:
Relabelled the switch after live testing. `show_ime_with_hard_keyboard` only overrides Android's suppression of the on-screen keyboard while a hardware keyboard is attached, so on an AVD without one the keyboard shows whatever the switch says, which reads as a broken control. The row is now **Force on-screen keyboard**, and the description says "Keeps the on-screen keyboard visible while a hardware keyboard is attached. A device with no hardware keyboard always shows it." Commit `58fa1b9`; the four links above are restacked.

PR #90, opened 2026-09-09
feat(hub): add a Location inspector section for Android and iOS

vonovak:
## Why

The Hub has no simulated-location control for either platform. serve-emu already
exposes `GET`/`POST /api/location`, and serve-sim's own preview tool drives
`xcrun simctl location`, so the whole gap is Hub-side.

One section serves both platforms. The two backends differ in exactly one way a
user can see: `adb emu geo fix` has no inverse, so only iOS can clear a fix. That
is the only asymmetry the types carry. Both backends agree on the other meaning,
"the last fix a backend confirmed applying", so the section never reads
`client.platform`.

## Scope

`DeviceClient` gains a location group placed after the camera group: `location`
(`DeviceGeoFix | null`), `locationPending`, `locationError`, `setLocation`, and
`clearLocation`. `DeviceCapabilities` gains `location: DeviceLocationCapabilities`,
which is `false` where no backend offers the control and carries `clear` only
where the fix can be removed.

Three new hub-client modules, none of them exported from the package index.
`android-location.ts` reads and writes `/api/location` through the Hub proxy with
an injected `fetch`. `ios-location.ts` runs `xcrun simctl location <udid> set|clear`
over the existing exec-ws host channel and rejects on a non-zero exit, because
that channel resolves failures. `useDeviceLocation.ts` holds the pending, error,
and generation bookkeeping once and derives the capability from the
`DeviceLocationBackend` record it is given, so the flag cannot drift from the
method.

In hub-components, `LocationSection.tsx` renders between Camera and Events,
`geoFixInput.ts` holds the parsing, preset table, and formatting as pure
functions, and `SidebarTextInput.tsx` is the first free-text control in the
inspector. `hub-components` still imports no value from `hub-client`.

Out of scope, deliberately. Route playback (`/api/route`, `/api/route/control`).
The `altitude`, `satellites`, and `velocity` fields that `POST /api/location`
accepts and this section does not send.

## Tradeoffs

serve-emu's `GET /api/location` returns the fix serve-emu itself last applied, not
a read of the emulator, so a fresh session reports `null` even when the emulator
holds a fix. iOS cannot report a fix at all. Rather than model that asymmetry in
the types, both are described by one honest sentence, and the section says "No fix
set in this session." until a write is confirmed.

An unanswered read is retried every 3 s rather than treated as "unsupported",
because the Hub can select a device before its backend is up. Only an answered
read turns support off. `android-camera.ts` already states that rule.

The coordinate boxes hold a draft, not the applied fix. They keep what you typed
after a Clear or after a backend normalises a value; the closing note is the
applied fix.

## Blast Radius

`DeviceCapabilities` gains a required field, so every `DeviceClient` implementation
and both hand-written test fixtures had to set it. The three real clients set it
from `useDeviceLocation`; the noop client and both fixtures set `false`, which
leaves every existing section-order assertion untouched.

Shared files are touched additively: `types.ts`, the three client hooks,
`LogSidebar.tsx`, `hub-components/src/index.ts`, and `inspectorSections.test.tsx`.
`SectionNote` gains an optional `id`. No existing behaviour changes.

## Verification

Gates, from the worktree root, after `bunx turbo run build --filter='./packages/@expo/hub-*'`.
`bun run lint` clean. `bun run typecheck` 11 tasks green.
`bun run test` 0 failures across hub-client (195), expo-device-hub (175), and
hub-components (53). `cd packages/expo-device-hub && bunx tsc --noEmit` prints no
non-vendor error. `bunx changeset status` bumps `expo-device-hub` only.

End to end in Chrome through Playwright, against the real shared devices, on the
final commit.

Android. Set 35.6812, 139.7671 in the Hub. `GET /api/location` through the Hub
proxy went from `location: null` to that exact fix. `adb shell dumpsys location`
independently reported `Location[gps 35.681198,139.767098]`, which is the same fix
at the emulator's quantization. A browser reload re-read it from serve-emu rather
than from client state. The section offers no Clear.

iOS. Set 48.8584, 2.2945 in the Hub. `simctl location` has no read, so the readback
is the guest's own `locationd`, which logged
`notifying clients ... at location (<+48.85840000,+2.29450000>)` three times right
after the write. Clear produced a `received daemon-side request to stop location
simulation` line and the note returned to "No fix set in this session.". iOS cannot
report a current fix, so the iOS proof is that log, not a queried value.

Two adversarial review rounds ran on the diff and found nine defects worth fixing;
commits 4 and 5 fix them and the e2e was re-run after each.


vonovak:

- Gates rerun in a fresh worktree at `9a818cd`: `bun run lint`, `bun run typecheck`, `bun run test --force` exit 0; `tsc --noEmit` in `packages/expo-device-hub` shows no error outside `vendor/`; `changeset status` bumps `expo-device-hub` only.
- Receipts: `useDeviceLocation` holds one fix per backend record; the Android and iOS wire modules inject `fetch` and `exec`; the section uses the shared primitives; the README gains one bullet.
- Runtime, independent rerun from the verifier worktree on port 8101: Android `GET /api/location` through the Hub proxy went from `location: null` to `35.6812, 139.7671` after Set location, and `dumpsys location` reported `Location[gps 35.681198,139.767098]`. iOS `locationd` logged three `notifying clients` lines for the set fix and a simulation stop after Clear.
- Rebased onto `vonovak/android-keyboard` (link 3); the verdict carries over.

vonovak:
Restack note: rebasing onto link 3 conflicted in six files where links 2 (accessibility) and 4 (location) both extend `DeviceClient`, `DeviceCapabilities`, the three client implementations, and two test fixtures. Every hunk was add/add, so the root kept both sides and added one commit that carries both links' fields in every fixture. Hub lint, typecheck, and the full test run pass on the rebased head. The Location section renders after Accessibility and before Events.

packages/@expo/hub-components/src/dashboard/LogSidebar.tsx
```diff
@@ -131,6 +132,8 @@ export function LogSidebar({
         )}
         {client?.capabilities.camera && <CameraSection client={client} />}
         {client?.capabilities.accessibility && <AccessibilitySection client={client} />}
+        {client?.capabilities.location && <LocationSection client={client} />}
```
krystofwoldrich-agent: **[critical]** `LocationSection` is rendered twice. Commit 71c0e63 added the `key={device?.id}` line below but kept this one. Every client with `capabilities.location` shows two Location sections, each with its own draft and error state, and each able to post a write. The order test in `inspectorSections.test.tsx` uses `indexOf`, so it passes with two sections. Delete this line, and add an assertion that the sidebar markup contains exactly one `<section aria-label="Location"`.

packages/expo-device-hub/src/dashboard/__tests__/inspectorSections.test.tsx
```diff
@@ -1323,7 +1359,9 @@ test('shows only the viewer-local frame option while iOS device settings are una
       events: true,
       camera: false,
       accessibility: false,
+      location: false,
       streamSettings: false,
+      location: false,
```
krystofwoldrich-agent: **[design]** Duplicate `location: false` key: commit 73facea added the same key at line 1362. `tsc --noEmit` in `packages/expo-device-hub` reports `TS1117: An object literal cannot have multiple properties with the same name` at (1364,7), which is the gate the PR description lists as clean. `bun test` still passes because JavaScript keeps the last key. Delete one of the two lines.

packages/@expo/hub-client/src/useDeviceLocation.ts
```diff
@@ -0,0 +1,123 @@
...
+        (result) => {
+          if (cancelled || generationRef.current !== generation || !result) return;
+          clearInterval(retry);
+          setState((current) => ({ ...current, ...result }));
+        },
+        () => {},
+      );
+    };
+    attempt();
+    const retry = setInterval(attempt, READ_RETRY_MS);
```
krystofwoldrich-agent: **[suggestion]** `setInterval(attempt, READ_RETRY_MS)` starts a new `read()` every 3 s whether or not the previous one has settled. A proxy that holds the request open for longer than 3 s stacks requests. `useAndroidCamera` guards the same poll with a `polling` flag and aborts the in-flight request on cleanup. Reuse that guard, or schedule the next attempt from the previous one's settlement.

krystofwoldrich-agent (changes requested):

**Verdict:** REQUEST_CHANGES
**Findings:** 1 critical · 1 design · 1 suggestion · 0 nit

The PR adds a `location` group to `DeviceClient`, two backend modules (`android-location.ts` over `/api/location`, `ios-location.ts` over `xcrun simctl location` on exec-ws), one shared hook (`useDeviceLocation`), and a `LocationSection` with pure parsing helpers. The design matches the existing camera pattern and the wire shapes match serve-emu's `location.ts` route. Two defects came in with the last two fix-up commits: `LogSidebar` renders `LocationSection` twice, and a test fixture has a duplicate object key that fails `tsc`. Both are one-line deletions.

- **[critical]** `packages/@expo/hub-components/src/dashboard/LogSidebar.tsx:135` — `LocationSection` is rendered twice (line 135 and the keyed line 136).
- **[design]** `packages/expo-device-hub/src/dashboard/__tests__/inspectorSections.test.tsx:1364` — duplicate `location: false` key; `tsc --noEmit` reports TS1117.
- **[suggestion]** `packages/@expo/hub-client/src/useDeviceLocation.ts:67` — the 3 s read retry has no in-flight guard.

Checks: `bun test` in hub-client 233 pass (1 file errored on a missing local package), hub-components 25 pass (7 files errored on a missing local package), `oxlint` clean in both. `tsc --noEmit` in `expo-device-hub` reports the TS1117 above.

vonovak:
Interrogate fix: two reviewers found the sidebar rendered the Location section twice on the stacked tip. The restack union had kept both the original line and the keyed replacement from `71c0e63`. Commit `a3504ff` keeps only the keyed line and the section test now asserts exactly one Location section. Links 5 to 7 are rebased on top.

vonovak:

**Resolved on the current head**
- Two `LocationSection` renders in `LogSidebar` (all three interrogate reviewers and the automated review). Fixed in `a3504ff`; the section test now asserts exactly one Location section, and the location e2e was rerun against the stacked tip.
- Duplicate `location: false` key in the `inspectorSections` fixture (automated review at `73facea`). Removed by a later reconciliation commit; `tsc --noEmit` in `packages/expo-device-hub` shows no error outside `vendor/` and the pre-existing `serve-emu-options.test.ts` errors on `main`.

**Act on**
- `useDeviceLocation` retries the read with `setInterval` every 3 s regardless of whether the previous read settled, so a slow proxy stacks requests. Reuse the `polling` flag and abort-on-cleanup pattern from `useAndroidCamera`, or schedule the next attempt from the previous settlement. (automated review)

**Consider**
- `useDeviceLocation` has no hook test although two of this link's commits fixed hook behaviour. `react-test-renderer` is already a dev dependency of the package. (interrogate B and the automated review)
- The location fetch carries no `AbortController`; a cancelled flag prevents the state update but the request runs to completion. (disclosed in the PR body)

vonovak:
Fix round on the triage above. New head `8259db4`.

**Fixed**

- The `setInterval` retry stacking reads (Act on). `6a25ff0` adds the failing test and puts the signal in `DeviceLocationBackend.read`; `e2148fe` gates the tick on a `reading` flag and aborts the outstanding read from the effect cleanup, following `useAndroidCamera`. `readAndroidLocation` now forwards the signal to `fetch`.
- The hook had no test (Consider) and the read carried no `AbortController` (Consider). Both fall out of the same fix. `src/__tests__/device-location.test.tsx` stubs `setInterval` to drive ticks and asserts three things: a tick during an in-flight read starts no second read, a settled read lets the next tick read again, and unmount aborts the signal. Two of the three fail on `a3504ff` (`Expected length: 1, Received length: 4`, and `signals[0]` undefined).
- The duplicate `location: false` key in the iOS `inspectorSections` fixture. The triage listed this as already resolved, but `tsc` still reported `TS1117` at `a3504ff`. `8259db4` removes it.

**Gates at `8259db4`**

`bun run lint`, `bun run typecheck`, `bun run test --force` exit 0 (1582 tests, 0 fail). `bunx tsc --noEmit` in `packages/expo-device-hub` now shows only the three pre-existing `serve-emu-options.test.ts` errors from `main`. `bunx changeset status` bumps `expo-device-hub` only. serve-emu is untouched.

**E2E rerun** under the shared lane on this head: Android sets 35.6812/139.7671, the proxy readback and `dumpsys` GPS fixes both agree, and a page reload re-reads and still shows the fix. iOS sets 48.8584/2.2945 with two matching `locationd` simulation lines, and Clear returns the section to "No fix set in this session." No page errors.

**Not done**

The comment-review pass proposed deleting four pre-existing doc comments and renaming `DeviceLocationRead.supported` and the `attempt` closure. All of that sits outside this fix round's diff, so I reverted it rather than churn reviewed code.

The request-stacking itself is proven by the unit test, not by the browser. It needs a proxy slow enough to outlast a 3 s tick; in the lane the read answers on the first attempt, so the e2e shows the read path unbroken rather than the stacking gone.

https://claude.ai/code/session_0111HkWoDi9TeKMVfAAp9z1H

vonovak:
## Fix round and restack

The act-on item landed at `b3a4bf4`: the read retry keeps one request in flight and aborts it on cleanup, with the signal threaded through `DeviceLocationBackend.read`. A `react-test-renderer` test fails on the old code with four stacked reads and an unmounted-abort error. The duplicate `location` capability key in the iOS fixture is gone, folded into the commit that introduced it.

krystofwoldrich-agent (approved):
[nit / follow-up] We could explore using expo-location to convert an address into latitude and longitude, so users can enter an address when setting the simulated location.

PR #91, opened 2026-09-10
feat(hub): stream Android activity metrics from serve-emu GET /api/metrics

vonovak:
## Why

The Hub shows CPU, memory, and network charts for the foreground iOS app. Android set `activity: null` and `capabilities.activity: false`, so the Current app section stopped at the App ID rows. serve-emu had no sampler and no metrics route. This PR adds both. The wire grammar stays identical to serve-sim `/metrics`, so the Hub reuses the existing `activity.ts` parsers and `ActivityCharts` without a change to either.

## Scope

serve-emu (in-repo since #78, so these are plain commits on this branch):

- New `src/metrics.ts`. `parseMetricsProbe(stdout)` parses one batched `adb shell` script (`/proc/stat`, the resumed activity's package and pid, `/proc/<pid>/stat`, `VmRSS`, `/proc/net/dev`) split on `---name` sentinel lines. `deriveMetricSample(prev, probe, t)` turns jiffy and byte deltas into a `MetricSample`. `MetricsSampler.subscribe(signal)` returns the SSE `Response`: a `:` comment, `event: meta`, `data:` frames every second, and a 15 s heartbeat. Each tick is one `adb shell` on `lane: "background"` with a 4 s deadline, and the next tick is scheduled only after the previous one settles.
- `DeviceSessionState` gains `readonly metrics: MetricsSampler` beside `logcat`, closed in `#dispose`. `ActiveDeviceSession` gains a `metrics` getter. One sampler per serial by construction.
- `GET /api/metrics` in all three chains: `middleware.ts`, `server.ts`, and the typed table (`inspection.ts`, `openMetrics` in `ApiDependencies`). `MetricsMeta` and `MetricSample` plus the `/api/metrics` contract and its `unsupportedStreamingResponse` parser in `shared/api-contracts.ts`.
- Route counts move to 57 method/path pairs across 40 paths in `tests/api-routes.test.ts` and `tests/api-contracts.test.ts`. `tests/server-request-gates.test.ts` gains the `POST /api/metrics` method gate. README `### Inspection` curl block and a paragraph, `docs:sync`, CHANGELOG.

Hub:

- New `hub-client/src/android-activity.ts`. `parseAndroidActivityFrame(eventType, data)` classifies one `EventSource` frame as `meta` or `sample` and drops anything else. `nextAndroidActivityAfterSilence(activity, clock)` decides the watchdog transition, so the hook holds no branching of its own.
- `useAndroidDevice` opens `EventSource(deviceApiUrl(baseUrl, '/api/metrics', device))`, appends samples with `appendActivitySample`, returns `activity`, and sets `capabilities.activity: true`. It marks `stale` after 8 s without a sample, `errored` on transport error, and `errored` when 8 s pass from connect with no sample at all.
- `inspectorSections.test.tsx` now expects `activity-charts` in the Android Current app section. Changeset for `expo-device-hub`, minor.

Out of scope: iOS code, `activity.ts`, and `ActivityCharts` are untouched.

## Tradeoffs

- Network rates are device-wide, from `/proc/net/dev` minus `lo`. This emulator has no per-uid counters: `xt_qtaguid` and `uid_stat` are absent and `/sys/fs/bpf` is denied. The README says so.
- `hostCores` carries the guest core count from `/proc/stat`, so it is known only after the first probe. A stream therefore writes `event: meta` on its first sample when the sampler had no probe yet. The first tick runs at once, so the delay is one probe round trip, not one interval.
- `cpuPct` is percent of one guest core from jiffy deltas, so neither `CLK_TCK` nor a wall clock enters the ratio. The divisor sums `user` through `steal` and skips `guest` and `guest_nice`, which `user` and `nice` already include. The Hub's existing copy reads "The host has N cores"; for Android that N is the guest count.
- CPU and memory describe the process that `pidof <package>` resolves. An app that declares `android:process` for its activity reports zero, and a multi-process app such as Chrome excludes its renderer children. The README says so.
- The sampler is a field on `DeviceSessionState` rather than a module-level cache keyed by serial, which is serve-sim's model. The session state is already the per-serial lifetime, so ref-counting and teardown on the device abort signal come for free.

## Blast Radius

- serve-emu: no existing route or type changes shape. The sampler adds one `adb shell` per second per subscribed device on the background lane. With no subscriber there is no adb traffic at all. The 4 s deadline and the background lane keep the tick off the interactive input path, and a failed tick is skipped rather than propagated.
- Hub: only the Android client changes. The Hub subscribes while the device is active, so an open Android device costs one background `adb shell` per second even when the Current app section is collapsed. iOS, the noop client, and the shared components are untouched. If the stream fails, the charts fall back to the existing message and the rest of the section is unaffected.
- Shared files (`middleware.ts`, `server.ts`, `api-contracts.ts`, `api/routes/inspection.ts`, `useAndroidDevice.ts`, `inspectorSections.test.tsx`) take additive blocks only, so sibling branches restack without conflicts.

## Verification

- serve-emu, in `packages/serve-emu/packages/serve-emu`: `bun run typecheck` (three tsconfigs, including tests) passes; `bun test` reports 931 pass, 0 fail; `bun run docs:check` reports the package README synchronized. New `tests/metrics.test.ts` covers parser fixtures captured from the real emulator, the delta math, the ref-counted sampler lifecycle through the SSE body, the 409 and 499 gates, and failed-probe tolerance. Separate tests use a deferred exec to prove that probes never overlap, that a subscriber joining mid-probe reuses the in-flight tick rather than starting a second loop, and that `close()` during a pending probe stops the loop. Every exec call is recorded with its lane and asserted to be `background` after the stream drains, so a wrong lane fails with a message instead of a timeout.
- Hub, from the repo root: `bun run lint`, `bun run typecheck`, and `bunx turbo run test --force` pass (185 hub-client, 24 hub-components, 174 expo-device-hub, 121 hub-android-utils, 44 hub-apple-utils, 931 serve-emu). `bunx tsc --noEmit` in `packages/expo-device-hub` reports 208 errors, all under `vendor/`, and zero outside it.
- `bunx changeset status`: `expo-device-hub` bumped at minor.
- End to end on `emulator-5554` through the Hub dev server on port 8095, driven by Playwright with headless Chrome under the shared-device lane. The driver asserts and fails loudly; it reported `ok: true` with no failures. Idle on the launcher: max CPU 0%, memory 91 MB. After launching Chrome the App ID flips to `com.android.chrome`, CPU peaks at 59%, and memory climbs to 214 MB. An independent `adb` readback computed from `/proc` outside the Hub during the same window returned 4 cores and 19% CPU for Chrome, and the chart's own tooltip text names the same core count. Curling the route through the Hub proxy returned an `event: meta` frame with `hostCores: 4` and 12 `data:` frames in 12 s. The browser console logged no errors.


vonovak:

- Gates rerun in a fresh worktree at `ea101c1`: Hub `bun run lint`, `bun run typecheck`, `bun run test --force` exit 0; `tsc --noEmit` in `packages/expo-device-hub` shows no error outside `vendor/`; serve-emu `bun run typecheck` 0, `bun test` 931 pass 0 fail, `docs:check` 0; the built `vendor/serve-emu/dist/middleware.js` serves `/api/metrics`; `changeset status` bumps `expo-device-hub` only.
- Receipts: `metrics.ts` runs one `adb shell` probe per tick with pure `parseMetricsProbe` and `deriveMetricSample`; the ref-counted `MetricsSampler` lives on `DeviceSessionState`; the SSE grammar matches serve-sim; the Hub subscribes with `EventSource`, reuses the `activity.ts` parsers, and the silence watchdog is a pure function.
- Runtime, independent rerun from the verifier worktree on port 8102: idle launcher reads 0% CPU and 93 MB; Chrome in the foreground reads 47% CPU and 144 MB with the App ID flipping to `com.android.chrome`; SSE frames flow through the Hub proxy after `event: meta`.
- Rebased onto `vonovak/location-section` (link 4); the verdict carries over.

vonovak:
Restack note: rebasing onto link 4 conflicted in the serve-emu route-count assertions (links 3 and 5 each bumped them by their own delta) and in one import block of `useAndroidDevice.ts`. The root kept both sides and added one commit that sets the stacked totals: 59 method/path pairs across 41 paths in `api-routes.test.ts` and 41 paths in `api-contracts.test.ts`. serve-emu tests (950) and the full Hub gates pass on the rebased head.

packages/serve-emu/packages/serve-emu/src/metrics.ts
```diff
@@ -0,0 +1,325 @@
...
+const METRICS_SAMPLE_INTERVAL_MS = 1_000;
+const METRICS_PROBE_TIMEOUT_MS = 4_000;
+const METRICS_HEARTBEAT_MS = 15_000;
+
+// Network is device-wide: this emulator exposes no per-uid counters
+// (`xt_qtaguid` and `uid_stat` are absent, `/sys/fs/bpf` is denied).
+const METRICS_PROBE_SCRIPT = [
+  "echo ---stat; head -n1 /proc/stat; grep -c '^cpu[0-9]' /proc/stat",
+  "echo ---fg; C=$(dumpsys activity activities 2>/dev/null | grep -m1 -E 'ResumedActivity[:=]'); echo \"$C\"",
+  "PKG=$(printf %s \"$C\" | sed -n 's/.*ActivityRecord{[^ ]* [^ ]* \\([A-Za-z0-9_.]*\\)\\/.*/\\1/p')",
```
krystofwoldrich-agent: **[suggestion]** Foreground-app detection is duplicated here and diverges from `/api/foreground`. `app-info.ts:foregroundComponent` reads `dumpsys window` first (`mCurrentFocus`, `mFocusedApp`, `mInputMethodTarget`) and falls back to `dumpsys activity activities` with regexes that accept any tokens before the component (`[^}]*\s`). This script reads only `dumpsys activity activities`, takes the first `ResumedActivity[:=]` line, and the `sed` pattern requires exactly two tokens before the package (`{[^ ]* [^ ]* `). When the two paths disagree, the App ID row and the chart `bundleId` differ, and `appendActivitySample` resets the chart history on each bundle change. The script already echoes the raw `$C` line under `---fg`, so `parseMetricsProbe` could apply the `app-info.ts` regexes to that line in TypeScript and keep `sed` only for the `pidof` argument, or `app-info.ts` could export the pattern list so both paths share it.

packages/serve-emu/packages/serve-emu/src/metrics.ts
```diff
@@ -0,0 +1,325 @@
...
+    if (this.#cores === null) return null;
+    return {
+      schemaVersion: 1,
+      udid: this.serial,
+      hostCores: this.#cores,
+      sampleIntervalMs: this.#intervalMs,
+    };
+  }
+
+  subscribe(signal?: AbortSignal): Response {
```
krystofwoldrich-agent, line 246: **[suggestion]** No subscriber cap. `LogcatHub.subscribe` rejects the ninth subscriber with a structured `logcat-subscriber-limit` error (`DEFAULT_MAX_LOGCAT_SUBSCRIBERS = 8`). This accepts any number of subscribers, each with its own heartbeat interval and stream. The per-tick cost is shared, so this is cheaper than logcat, but the package guidance asks to keep API inputs bounded. A small cap with the same error shape would match the existing pattern.

packages/serve-emu/packages/serve-emu/src/server.ts
```diff
@@ -2784,6 +2784,18 @@ export async function startServer(
         }
       }
 
+      if (url.pathname === "/api/metrics") {
+        if (req.method !== "GET")
+          return new Response("method not allowed", { status: 405 });
+        try {
+          sessions.assertCurrent(requestContext);
+          srv.timeout(req, 0);
+          return requestContext.metrics.subscribe(req.signal);
```
krystofwoldrich-agent, line 2844: **[suggestion]** `/health` has no metrics snapshot. `health()` includes `logcat: context.logcat.snapshot()` here (line 726) and in `middleware.ts:484`. The sampler exposes `subscriberCount`, `running`, and `meta`, but nothing surfaces them. The package guidance names `/health` as the machine-readable snapshot for agents. A `metrics: { subscribers, running, cores }` entry would show whether a device is being probed once per second.

packages/serve-emu/packages/serve-emu/tests/metrics.test.ts
```diff
@@ -0,0 +1,355 @@
...
+    expect(instance.subscriberCount).toBe(2);
+
+    a.abort();
+    const before = calls.length;
+    await Bun.sleep(40);
+    const during = calls.length - before;
+    b.abort();
+    await Bun.sleep(20);
+    expect(during).toBeGreaterThan(0);
+    expect(during).toBeLessThanOrEqual(10);
```
krystofwoldrich-agent, line 331: **[suggestion]** This assertion depends on scheduling: `during > 0 && during <= 10` after a 40 ms `Bun.sleep` with a 5 ms interval. `LogcatHub` takes a `LogcatClock` and `RoutePlayback` takes `routeClock`, so their tests drive timers deterministically. `MetricsSampler` takes `now` but calls the global `setTimeout` and `setInterval`. An injected clock would remove the sleeps here and in the other lifecycle tests, and follow the existing pattern.

packages/serve-emu/packages/serve-emu/src/middleware.ts
```diff
@@ -1539,6 +1539,21 @@ async function createAppInternal(
       return logcatStream(req, url);
     }
 
+    if (url.pathname === "/api/metrics") {
+      if (req.method !== "GET") return new Response("method not allowed", { status: 405 });
+      if (!isAllowedBrowserOrigin(req, opts)) {
+        return Response.json({ error: "forbidden_origin" }, { status: 403 });
+      }
+      const response = deviceState.metrics.subscribe(req.signal);
+      const headers = new Headers(response.headers);
```
krystofwoldrich-agent, line 1570: **[nit]** Lines 1548 to 1554 repeat lines 686 to 693 of `logcatStream` (copy headers, apply `corsHeadersForRequest`, rebuild the `Response`). A small `withCors(req, response)` helper would serve both routes.

krystofwoldrich-agent (commented):

**Verdict:** APPROVE
**Findings:** 0 critical · 0 design · 4 suggestion · 3 nit

PR 5/7 in the stack. The PR adds a `MetricsSampler` to serve-emu (one batched `adb shell` per second on the background lane, ref-counted by SSE subscriber, closed with the device session), mounts `GET /api/metrics` in the standalone server, the middleware, and the typed route table, and makes `useAndroidDevice` consume the stream with the existing `activity.ts` parsers. The wire format matches serve-sim `/metrics`, so the Hub reuses `ActivityCharts` unchanged. The probe script is a constant, so no request input reaches the shell. Lifecycle handling is covered by tests. The main concern is a second, independent foreground-app detector inside the probe script that can disagree with `/api/foreground`.

Checks run: serve-emu `typecheck` passes, `bun test` 950 pass / 0 fail, `docs:check` synchronized. hub-client `tsc --noEmit` passes, `bun test` 250 pass / 0 fail, `oxlint` clean. `inspectorSections.test.tsx` 47 pass.

**Findings**
- [suggestion] `packages/serve-emu/packages/serve-emu/src/metrics.ts:13` — foreground-app detection duplicates and diverges from `app-info.ts`; the App ID row and chart `bundleId` can disagree.
- [suggestion] `packages/serve-emu/packages/serve-emu/src/metrics.ts:201` — no subscriber cap, unlike `LogcatHub` (8).
- [suggestion] `packages/serve-emu/packages/serve-emu/src/server.ts:2793` — `/health` has no metrics snapshot; logcat has one.
- [suggestion] `packages/serve-emu/packages/serve-emu/tests/metrics.test.ts:252` — timers are not injectable, so tests rely on wall-clock sleeps.
- [nit] `packages/serve-emu/packages/serve-emu/src/middleware.ts:1548` — CORS re-wrap duplicated from `logcatStream`.
- [nit] `packages/serve-emu/packages/serve-emu/src/device-session-state.ts:75` — `metrics` not injectable through options, unlike `recorder` and `logcat`.
- [nit] `packages/@expo/hub-client/src/useAndroidDevice.ts:1541` — Hub copy says "host" cores; Android reports guest cores (follow-up in hub-components).

packages/serve-emu/packages/serve-emu/src/device-session-state.ts
```diff
@@ -70,6 +72,7 @@ export class DeviceSessionState {
     this.#now = options.now ?? Date.now;
     this.recorder = options.recorder ?? new SessionRecorder();
     this.logcat = options.logcat ?? new LogcatHub(options.serial);
+    this.metrics = new MetricsSampler({ serial: options.serial });
```
krystofwoldrich-agent, line 75: **[nit]** `recorder` and `logcat` can be passed in through `DeviceSessionStateOptions`; `metrics` is always constructed here with the real `execText`. A `metrics?: MetricsSampler` option would keep the three fields symmetric and let session-state tests substitute a fake exec.

packages/@expo/hub-client/src/useAndroidDevice.ts
```diff
@@ -1509,6 +1518,54 @@ export function useAndroidDeviceClient(options: DeviceConnectionOptions): Device
...
+    } catch {
+      setActivity({ ...EMPTY_ANDROID_ACTIVITY, errored: true });
+      return;
+    }
+    const onFrame = (event: Event) => {
+      const frame = parseAndroidActivityFrame(event.type, String((event as MessageEvent).data));
+      if (!frame) return;
+      if (frame.kind === 'meta') {
+        setActivity((current) =>
+          current ? { ...current, hostCores: frame.hostCores, errored: false } : current,
```
krystofwoldrich-agent, line 1578: **[nit]** `ActivitySection` renders "The host has N cores" from `hostCores`. For Android, this value is the emulator guest core count from `/proc/stat`. The PR body notes this. The copy lives in `@expo/hub-components`, outside this diff, so this is a follow-up rather than a change here.

vonovak:

vonovak:

**Act on**
- Foreground-app detection in `metrics.ts` (one `grep` on `ResumedActivity` plus one `sed`) diverges from `app-info.ts` (three ordered regexes over `dumpsys window`, then `dumpsys activity`). When they disagree, `appendActivitySample` clears the chart while the Current app header names another app. Reuse one detector, or feed the package from `getForegroundApp` into the sampler. (interrogate reviewer B and the automated review, independently)
- No subscriber cap on the sampler; `LogcatHub` rejects the ninth subscriber with a structured error. Add the same bound and error shape. (automated review)

**Consider**
- `/health` has no metrics snapshot (`subscribers`, `running`, `cores`) although the package guidance names `/health` as the agent-facing snapshot. (automated review)
- `metrics.test.ts:252` depends on real timers (`during > 0 && during <= 10` after a 40 ms sleep). `MetricsSampler` takes `now` but not a clock; inject one like `LogcatHub` and `RoutePlayback` do. (automated review)
- Derive the serve-emu route-count assertions from `EXPECTED_ROUTES.length` so the next route never needs a reconciliation commit; two such commits exist in this stack only to re-sum counts. (all three interrogate reviewers)

**Noted**
- `MetricsMeta.udid` and `MetricSample.bundleId` carry an Android serial and package name so the shared Hub parser works unchanged; the "host has N cores" copy shows the guest core count. Disclosed in the PR body. (interrogate B, automated review nit)
- The CORS header block repeats `logcatStream`'s; `metrics` is the only session field not injectable through `DeviceSessionStateOptions`. (automated review nits)

vonovak:
Fix round for the two **Act on** items. New head `48449af`, five commits on top of `3bde36d`.

**One foreground-app detector.** `e399787` moves the dumpsys patterns out of `app-info.ts` into `src/foreground-component.ts`, which exports `parseForegroundWindowDump` and `parseForegroundActivityDump` plus the grep alternation the probe sends to the device. The alternation is `DETECTORS.map(({ key }) => key).join("|")`, so a new detector cannot drift from the filter, and a test asserts every line a detector reads survives that filter. `50bde5c` puts the sampler on it: the script echoes the window and activity lines the detector reads, collects `CANDIDATES` from them, and emits `pid`, `stat` and `rss` per candidate; `parseMetricsProbe` picks the one the detector chose. The tick still costs one `adb shell`, which is why the device collects every candidate instead of choosing one, since `pidof` needs a name the host has yet to parse.

Three ways the old pipeline disagreed with `/api/foreground`, each now a unit test and each reproduced before the fix (`evidence/android-activity/divergence-check.txt`): it never read `dumpsys window`, so a focused app lost to whatever `dumpsys activity activities` resumed (`com.android.launcher3` vs `com.example.focused`); `grep -m1` took the first resumed line rather than ranking `topResumedActivity=` first (`com.example.earlier` vs `com.example.top`); and its `sed` accepted a component the shared patterns reject (`com.example.app` vs none).

**Subscriber cap.** `635f28e` adds `DEFAULT_MAX_METRICS_SUBSCRIBERS = 8` and refuses the ninth stream with `{ok: false, code: "metrics-subscriber-limit", error: "metrics subscriber limit is 8"}` at HTTP 429, matching `LogcatHub`. I did not add a `positiveInteger` guard on the new option: `intervalMs` has none either, and `logcat.ts` keeps that helper private.

**Also.** `97a07d7` deletes a duplicate `location: false` key that the restack left in an `inspectorSections.test.tsx` fixture; it was the only `tsc --noEmit` error in this link outside the three that come from main. `48449af` is the `/no-comments` pass: `parseProcSection` now returns stats keyed by package instead of latching on the target, which let two design comments go.

**Left alone.** The three **Consider** items stay out of scope: no `/health` metrics snapshot, no injected clock for `MetricsSampler`, and no `EXPECTED_ROUTES.length`-derived route counts. The last one spans links, so the root is better placed to make it.

**Verification.** serve-emu `typecheck`, 960 tests and `docs:check` pass; Hub `lint`, `typecheck` and `test --force` pass (250 + 180 + 121 + 960, 0 fail); `tsc --noEmit` in `packages/expo-device-hub` reports only the three known `serve-emu-options.test.ts` errors. E2E rerun under the lane at this head (`evidence/android-activity/summary-run-5.json`, `lane-run-5.log`), with three assertions added to the driver: the SSE `bundleId` equals `/api/foreground`'s `packageName` (`com.android.chrome` for both), `memBytes` is read off the device (160677888), and the cap refuses a stream with 429 and `metrics-subscriber-limit` (attempt 8, since the open page holds the first slot). CPU still rises 0 to 59 against a 12.1% independent adb readback.

The extra `dumpsys window` costs something, so I measured it on the emulator (`evidence/android-activity/probe-tick-cost.txt`): the replaced script ran at a 70 ms median, the current one at 96 ms, over five runs each. That is 26 ms more on a 1000 ms interval, against a 4000 ms probe timeout.

vonovak:
## Fix round and restack

Both act-on items landed at `de9782c`: the metrics probe and `/api/foreground` now share one detector in `foreground-component.ts`, so the chart and the Current app header cannot name different apps, and the sampler caps subscribers at eight with a `metrics-subscriber-limit` error matching `LogcatHub`. The probe still costs one `adb shell` per tick, measured at 70 to 96 ms against a 1000 ms interval. The e2e asserts the SSE `bundleId` equals `/api/foreground` and that the ninth stream is refused.

krystofwoldrich-agent (approved):
Note: On iOS, we currently only show these stats for user-installed apps, not the default preinstalled apps. I do not remember the reason for that restriction; it may simply be a technical limitation. Mentioning this for context on the difference between platforms, not as a requested change here.

PR #92, opened 2026-09-10
feat(hub): add a Permissions section for the foreground app

vonovak:
## Why

The Hub had no way to see or change an app's permissions. serve-emu offered grant only, and serve-sim had grant, revoke, and reset for 16 services but only through its CLI over exec-ws. A person who tests permission prompts needs to grant, revoke, and reset from the inspector and see the current state. This closes the "permissions" row of `ANDROID-PARITY.md`.

## Scope

serve-emu (in-repo):
- `src/app-permissions.ts` (new): `parseRuntimePermissions` reads the `runtime permissions:` block of `dumpsys package` and dedupes by name, since the block prints once per user. `listPermissions` returns `{ok, packageName, permissions[{name, granted, flags}]}`. `resetPermissions` is per package: one `adb shell` that runs, each command marking its own failure since the shell reports only the last status, `pm grant` or `pm revoke` per runtime permission by its `GRANTED_BY_DEFAULT` flag, `pm clear-permission-flags <pkg> <perm> user-set user-fixed`, then `appops reset <pkg>`. `pm reset-permissions` is device-wide on SDK 36, so it is not used.
- `src/app-management.ts`: `revokePermission` next to `grantPermission`; `adb()` is exported for the new module.
- Routes `GET /api/apps/permissions?packageName=`, `POST /api/apps/revoke`, `POST /api/apps/reset-permissions` on `middleware.ts`, `server.ts`, and `api/routes/applications.ts`; contracts and `parseAppPermissionsResponse` in `shared/api-contracts.ts`; route-count, request-gate, and contract tests; README `### Apps And Files`; CHANGELOG.

serve-sim (submodule pin to https://github.com/expo/serve-sim/pull/144): `GET {base}/permissions?device=&bundleId=` and `POST {base}/permissions {bundleId, id, action}` returning `{ok, bundleId, permissions[{id, state}]}`.

hub-client:
- `types.ts`: `AppPermission {id, label, state}`, `AppPermissionState = granted | denied | limited | undetermined`, `AppPermissionAction`; `DeviceClient` gains `permissions`, `permissionsPending`, `permissionsError`, `setPermission`, `resetPermissions`, `refreshPermissions`, and `capabilities.permissions`.
- `app-permissions.ts` (new): the `PermissionsBackend` boundary (`list`, `write`, `reset`, each resolving to the fresh list) and the pure merge helpers `stalePermissionIds` and `applyPermissionsRead`.
- `android-permissions.ts` and `ios-permissions.ts` (new): one backend factory per platform with its parser and label rule, `fetchImpl` injected.
- `useAppPermissions.ts` (new): one hook for both platforms. A list result applies only when the target it was issued for is still current; ids with a write in flight, or written after the read started, keep their current row. Pending ids come from `KeyedWriteTracker`. A successful read clears the error.
- `useAndroidDevice.ts`, `useIosDevice.ts`, `useNoopDeviceClient.ts`: wire the hook and the capability flag.

hub-components and Hub:
- `PermissionsSection.tsx` (new): rows of label and state with Grant and Revoke, a Reset all button, notes for no foreground app, first read, empty list, and error, and Retry when the first read fails. It reads the list when it opens and when the foreground app changes while open. Nothing reads on a timer.
- `LogSidebar.tsx` renders it after Camera when `capabilities.permissions` is set. `inspectorSections.test.tsx` gains the gating and order test. Changeset for `expo-device-hub`.

## Tradeoffs

- Reads happen on open and after each write only. The Android settings poll already spawns nine adb reads per device every three seconds, so a permissions poll would cost more than it returns. The cost is that a permission changed outside the Hub shows up on the next open or write.
- Android reset is a reconstruction of the manifest default from `GRANTED_BY_DEFAULT`, not the platform's own reset, because the platform reset is device-wide.
- `pm revoke` and the reset chain stop the app on Android. The foreground app can change after a revoke, and the section follows the foreground app, so the list may switch to the new app on the next foreground poll. The section does not pin the app it opened with.
- The iOS list is the serve-sim catalogue (16 entries) with the app's current state for each, not the set the app declares. Android lists only the runtime permissions in the package's dumpsys block. Filtering iOS by `Info.plist` usage strings is a follow-up.
- A serve-sim HTTP route instead of exec-ws CLI calls, so the Hub does not depend on the vendored CLI path or on parsing CLI JSON. serve-sim's writers are `execFileSync`, so a reset of all 16 entries blocks its event loop for about half a second.

## Blast Radius

New section, new routes, new client fields. Existing routes and sections are unchanged. Every `DeviceClient` fixture needs the six new fields and the capability flag, which the type system enforces. Shared files are edited additively so sibling branches restack without conflicts.

## Verification

- `bun run lint`, `bun run typecheck`, `bun run test` at the worktree root: green.
- `cd packages/expo-device-hub && bunx tsc --noEmit | grep -v '^vendor/'`: prints nothing.
- serve-emu in `packages/serve-emu/packages/serve-emu`: `bun run typecheck`, `bun test` (920 pass), `bun run docs:check` synchronized.
- serve-sim: `bunx tsc --noEmit` clean, permissions route and state tests pass (31).
- `bunx changeset status` lists `expo-device-hub` at minor.
- One `interrogate` reviewer (claude-opus-5) on the diff: acted on the error dead end, the reset that hid failures, the render-phase ref writes, the hanging serve-sim body read, and the hardcoded Android capability. Left open: hook tests through a client renderer, accessible names for the Grant and Revoke buttons, and the two tradeoffs above.
- End to end in the Hub website with Playwright under the shared device lane. Android: Chrome (`com.android.chrome`) in the foreground, `CAMERA` row read `Denied`; Grant flipped the row to `Granted` and `adb shell dumpsys package com.android.chrome` read `android.permission.CAMERA: granted=true`; Revoke flipped it back and dumpsys read `granted=false`. iOS: `dev.expo.Payments` in the foreground, `Camera` row read `Not asked`; Grant read `Granted` and `sqlite3 TCC.db` returned `kTCCServiceCamera|dev.expo.Payments|0|2`; Revoke read `Denied` and sqlite3 returned `auth_value` 0. The route's own GET and serve-sim's `listPermissions` from source agreed at each step. Both devices were left as found. Section screenshots before, granted, and revoked for both platforms are in the run evidence (`evidence/permissions/*-section-*.png`), kept locally by the program.


vonovak:

- Gates rerun in a fresh worktree at `1912e17` with the serve-sim submodule at `a08fd7e` built from the pushed branch: Hub `bun run lint`, `bun run typecheck`, `bun run test --force` exit 0; `tsc --noEmit` in `packages/expo-device-hub` shows no error outside `vendor/`; serve-emu `bun run typecheck` 0, `bun test` 0 fail, `docs:check` 0; serve-sim `tsc --noEmit` 0 and its permissions tests pass; the vendored serve-sim middleware carries the permissions route; `changeset status` bumps `expo-device-hub` only.
- Receipts: one `PermissionsBackend` per platform and one `useAppPermissions` hook over `KeyedWriteTracker`; serve-emu list, revoke, and reset beside `grantPermission` across the three chains; the serve-sim typed route is expo/serve-sim #144 and the pin points at it; shared-file edits are additive.
- Runtime, independent rerun from the verifier worktree on port 8103: Android Chrome CAMERA row went `Denied` to `Granted` to `Denied` while `dumpsys package` read `granted=true` then `granted=false`; iOS `dev.expo.Payments` camera went `Granted` to `Denied` while `TCC.db` read `auth_value` 2 then 0. Both devices were restored to their initial state.
- Rebased onto `vonovak/android-activity` (link 5); the verdict carries over.

vonovak:
Restack note: rebasing onto link 5 conflicted in the serve-emu CHANGELOG and route-count assertions (stacked totals are now 62 method/path pairs across 44 paths) and in the shared client files and fixtures where links 2, 4 and 6 all extend `DeviceClient` and `DeviceCapabilities`. Every hunk was add/add; the root kept both sides, dropped the duplicate `permissions` capability line the union produced, and added one commit that carries all three links' fields in every fixture. Hub lint, typecheck, and the full test run pass on the rebased head.

vonovak:
Correction to the restack note: the fixture-reconciliation commit had re-staged the assemble worktree's serve-sim pointer and moved the pin back to main's `55fec25`. A follow-up commit restores the pin to `a08fd7e` (expo/serve-sim #144). The verified runtime evidence was taken at `1912e17`, which already carried that pin; the stacked tip now matches it.

packages/serve-sim
```diff
@@ -1 +1 @@
-Subproject commit 55fec25295316e9757bf376c136275b59aea3df5
+Subproject commit a08fd7ef3d4a02e0c19902392a0026afc3130e6a
```
krystofwoldrich-agent: **[design]** `.gitmodules` tracks branch `expo-device-hub` for this submodule. The previous pin `55fec25` is on `origin/expo-device-hub`. This pin `a08fd7e` is only on `origin/vonovak/permissions` (serve-sim PR #144). Pins on `main` land through updates to the tracked branch (for example `bf38c74 Update serve-sim submodule to 55fec (#28)`). If serve-sim #144 is squash-merged or its branch is deleted, this pin points at a commit that is not on any tracked branch, and `bun run submodule:update` moves away from it. Merge serve-sim #144 into `expo-device-hub` first, then pin to that merged commit.

packages/serve-emu/packages/serve-emu/src/app-permissions.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/** Per package: `pm reset-permissions` is device-wide. */
+export async function resetPermissions(
+  serial: string,
+  packageNameValue: string,
+  dependencies: AppManagementDependencies = {},
+): Promise<AppActionResult> {
+  const pkg = packageName(packageNameValue);
+  const { permissions } = await listPermissions(serial, pkg, dependencies);
+  const commands = permissions.flatMap(({ name, flags }) => [
```
krystofwoldrich-agent: **[suggestion]** The chain emits `pm grant` or `pm revoke` for every entry in the runtime block, without checking `granted`. The README added in this PR states that `pm revoke` stops the app, so a reset on an app with several already-denied permissions stops it once per permission and adds one failure point per command. Skipping entries where `granted` already equals the `GRANTED_BY_DEFAULT` target would reduce both. The Verification section covers Grant and Revoke on one permission per platform and does not mention Reset all; an end-to-end reset on an app with several runtime permissions would confirm that every generated `pm` command succeeds.

packages/@expo/hub-client/src/useAppPermissions.ts
```diff
@@ -0,0 +1,125 @@
...
+
+interface UseAppPermissionsOptions {
+  active: boolean;
+  appId: string | null;
+  backend: PermissionsBackend | null;
+}
+
+type ListRequest = (backend: PermissionsBackend, appId: string) => Promise<AppPermission[]>;
+
+export function useAppPermissions({ active, backend, appId }: UseAppPermissionsOptions) {
```
krystofwoldrich-agent, line 21: **[suggestion]** The pure helpers and both backends are tested. The hook is not, and it holds the intricate part: the child-effect-before-parent-effect ordering that the commit history fixed twice, the drop of results for a stale `target`, the version bump before `request`, and `tracker.reset()` while a write is in flight. hub-client tests already use `react-test-renderer`, so a test that renders the hook, changes `appId` during an in-flight write, and asserts the pending set and the list would cover the fixed regressions.

packages/@expo/hub-components/src/dashboard/PermissionsSection.tsx
```diff
@@ -0,0 +1,95 @@
...
+  client,
+  defaultOpen = false,
+}: {
+  client: DeviceClient;
+  /** Whether the section is initially expanded. */
+  defaultOpen?: boolean;
+}) {
+  const [open, setOpen] = useState(defaultOpen);
+  const { permissions, permissionsPending, permissionsError, refreshPermissions } = client;
+  const appId = client.foregroundApp?.id ?? null;
```
krystofwoldrich-agent, line 27: **[suggestion]** `appId` is read from `client.foregroundApp` on every render. The PR body states that `pm revoke` stops the app on Android and that the list may switch to the new foreground app on the next poll. In that case the row the person just changed disappears right after the click, and the hook resets `permissions` to null for the new target. Is holding the last non-null `appId` while the section is open an option, with a note when it differs from the current foreground app? This is a question, not a blocker.

packages/@expo/hub-client/src/types.ts
```diff
@@ -571,6 +585,19 @@ export interface DeviceClient {
   /** Remove the simulated fix. A no-op unless `capabilities.location` carries `clear`. */
   clearLocation: () => void;
 
+  /** Permissions of the foreground app, or null while unknown or without a foreground app. */
+  permissions: readonly AppPermission[] | null;
+  /** Permission ids with a write in flight. A reset holds every id. */
+  permissionsPending: ReadonlySet<string>;
+  /** Last failed permission request, cleared when the next write starts. */
```
krystofwoldrich-agent, line 609: **[nit]** The doc says the error is cleared when the next write starts. `useAppPermissions.ts:64` also clears it on a successful read, and the Retry button in the section depends on that path. Add "or when a read succeeds".

packages/@expo/hub-client/src/android-permissions.ts
```diff
@@ -0,0 +1,66 @@
...
+export function parseAndroidPermissions(payload: unknown): AppPermission[] | null {
+  const data = asRecord(payload);
+  if (!data || data.ok !== true || !Array.isArray(data.permissions)) return null;
+  const permissions: AppPermission[] = [];
+  for (const entry of data.permissions) {
+    const item = asRecord(entry);
+    if (!item || typeof item.name !== "string" || typeof item.granted !== "boolean") return null;
+    permissions.push({
+      id: item.name,
+      label: humanize(item.name.startsWith(PREFIX) ? item.name.slice(PREFIX.length) : item.name),
```
krystofwoldrich-agent, line 23: **[nit]** `humanize("com.example.CUSTOM_THING")` yields `Com.example.custom thing`, and the test enshrines it. Using the segment after the last dot would give `Custom thing`, in line with the `android.permission.` rows.

krystofwoldrich-agent (changes requested):

**Verdict:** REQUEST_CHANGES
**Findings:** 0 critical · 1 design · 3 suggestion · 2 nit

The PR adds a Permissions section that lists, grants, revokes, and resets the foreground app's permissions. serve-emu gains three routes on all three route chains, with validated inputs and contract parsers that follow the existing `/api/apps/grant` shape. hub-client adds a `PermissionsBackend` boundary per platform and one `useAppPermissions` hook that drops results for a stale target and holds rows with a write in flight. The main concern is the serve-sim submodule pin: it points at a commit that exists only on the serve-sim PR branch, not on the branch `.gitmodules` tracks.

Checks run locally: `bun run typecheck` at the root exits 0. `bun test` passes in hub-client (261), hub-components (58), expo-device-hub (181), and serve-emu (956). serve-emu `bun run docs:check` reports the README synchronized.

Stack note: the duplicated `<LocationSection>` lines in `LogSidebar.tsx` come from the base branch (PR #91), not from this PR.

### Findings

- **[design]** `packages/serve-sim:1` — the pin `a08fd7e` is only on `origin/vonovak/permissions`; `.gitmodules` tracks `expo-device-hub`. Merge serve-sim #144 first, then pin to the merged commit.
- **[suggestion]** `packages/serve-emu/packages/serve-emu/src/app-permissions.ts:71` — the reset chain runs `pm grant`/`pm revoke` for every permission, including ones already at their default; each `pm revoke` stops the app. Reset all is not in the Verification section.
- **[suggestion]** `packages/@expo/hub-client/src/useAppPermissions.ts:21` — no hook-level test for the target-change and in-flight-write paths that the commit history fixed twice.
- **[suggestion]** `packages/@expo/hub-components/src/dashboard/PermissionsSection.tsx:27` — the section follows the foreground app, so an Android revoke can switch the list away right after the click.
- **[nit]** `packages/@expo/hub-client/src/types.ts:592` — `permissionsError` doc omits that a successful read clears it.
- **[nit]** `packages/@expo/hub-client/src/android-permissions.ts:23` — labels for names outside `android.permission.` keep the package prefix (`Com.example.custom thing`).

vonovak:

vonovak:

**Act on**
- Landing order: the serve-sim pin `a08fd7e` exists only on `origin/vonovak/permissions` (expo/serve-sim #144). If #144 is squash-merged or its branch deleted, the pin points at an unreachable commit and `bun run submodule:update` moves away from it. Merge #144 into `expo-device-hub` first (a merge that keeps the commit, or re-pin to the merged commit), then land this PR. (automated review)
- `capabilities.permissions` ignores `active`: `permissionsBackend` is memoized on `[baseUrl, targetDevice]` while `accessibilityLoader` two links earlier includes `active`. With `enabled: false` and a stale foreground app the section renders and sticks on "Reading permissions…" with no error path. Add `active` to the memo. (interrogate reviewer A)
- Reset issues `pm grant` or `pm revoke` for every runtime permission without checking its current state, and `pm revoke` stops the app, so a reset can stop it once per already-denied permission. Skip entries already at the default state, and add Reset all to the e2e. (automated review)
- `parseAppPermissionsResponse` has no test; only the keyboard parser was added to `api-contracts-complete.test.ts`. One valid and one rejecting case. (interrogate B)
- Commit history: the union commits inside this link (`c6db658` to the comment pass) do not typecheck until the reconciliation commit at the tip (duplicate `permissions` key in `useAndroidDevice.ts`). The PR head is green, but the link cannot be bisected. Fold the reconciliation hunks into the commits that introduced them with an autosquash rebase. (interrogate A)

**Ask**
- `appId` follows `client.foregroundApp` on every render. After `pm revoke` stops the app, the row the person just changed can disappear on the next poll and the list resets for the new foreground app. Hold the last non-null `appId` while the section is open, with a note when it differs from the current foreground app? Product call. (automated review)

**Consider**
- `useAppPermissions` has no hook test although it holds the effect-ordering fix, the stale-target guard, the version bump, and `tracker.reset()` during a write. `react-test-renderer` is already a dev dependency. (interrogate B and the automated review)

**Noted**
- The `permissionsError` doc should say "or when a read succeeds"; `humanize("com.example.CUSTOM_THING")` yields `Com.example.custom thing`. (automated review nits)

vonovak:
Fix round for the combined review triage. Three commits on top of `150330c`, head `ae02c44`.

**Fixed**
- `permissionsBackend` ignores `active`: `ae02c44` gates the memo on `active` in `useAndroidDevice.ts` and `useIosDevice.ts`. New `inactive-client-capabilities.test.tsx` renders both clients disabled, enabled, then disabled again and asserts `capabilities.permissions`. The Android assertion fails without the fix. The iOS assertion passes without its hunk because the config effect already clears `deviceUdid` when inactive; the hunk closes the one-render window and keeps the two clients alike.
- Reset issues `pm grant` or `pm revoke` for every permission: `da9d2c2` emits the state change only where `granted` differs from the `GRANTED_BY_DEFAULT` target and still clears `user-set user-fixed` on every entry. The reset test now covers an entry in each of the four states and fails without the fix. README reset text updated in both copies via `docs:sync`.
- `parseAppPermissionsResponse` has no test: `6215850` adds one valid case and two rejecting cases (`granted` not boolean, `permissions` not an array) to `api-contracts-complete.test.ts`.
- Reset all in the e2e: the driver now grants Camera on Chrome, clicks Reset all, polls `dumpsys` until CAMERA reads `granted=false` (602 ms), asserts no error alert, asserts every other runtime permission line is unchanged, relaunches Chrome, and asserts the Camera row reads Denied. Green on run 6 with 18 already-denied lines in the dump. One observation from that run: after `pm revoke` stops Chrome, the section switches to the launcher's rows before the reset returns, so the Camera row is not visible until Chrome is relaunched. That is the open Ask item about holding the last app id.

**Left as is**
- Landing order for the serve-sim pin and the commit-history fold belong to the root.
- Holding the last non-null `appId` is a product call (Ask). The e2e observation above is the concrete case for it.
- Consider and Noted items untouched.

Gates at `ae02c44`: Hub lint, typecheck, `test --force` (6 tasks) pass; `tsc --noEmit` outside `vendor/` shows only the three pre-existing `serve-emu-options.test.ts` errors; serve-emu typecheck, `bun test` (956 pass), `docs:check` pass; `changeset status` lists `expo-device-hub` only.

vonovak:
## Fix round and restack

Three act-on items landed at `1e1fb6b`: Reset skips permissions already at their default, so it stops the app once instead of once per already-denied permission; `permissionsBackend` is gated on `active` in both clients; and `parseAppPermissionsResponse` has valid and rejecting tests. The e2e gained a Reset all step, which exposed the concrete case behind the open question on this PR: after `pm revoke` stops the app, the foreground poll switches the section to the launcher before the reset returns. That question is still yours.

krystofwoldrich-agent (approved):
[nit] As a follow-up to this stack, we should update the sidebar action buttons to all use SidebarActionButton so their visuals are consistent. This includes Grant, Revoke, Reset all, Set location, and Choose PNG, matching Press, Shut down, and Remove.

vonovak:
Scoped this link to Android and dropped the serve-sim dependency.

serve-sim exposes permissions over its CLI channel rather than HTTP, so the iOS half needed a route in another repo (expo/serve-sim#144) and a submodule pin at a commit on that PR's branch. That coupled this stack to a second review queue, and would have broken outright if #79 lands first and turns serve-sim into plain files.

Removed: `ios-permissions.ts` and its test, the iOS backend wiring in `useIosDevice.ts`. The submodule pin is back to `main`'s `55fec25`. The shared `useAppPermissions` hook, `app-permissions.ts` and `PermissionsSection` are untouched, and iOS now reports `capabilities.permissions` false, so the section simply does not render there. No stubs and no platform branch.

The capability test now expects the real value per platform rather than true for both. The changeset and the README say Android.

Verified after the change: lint, typecheck and the full suite pass, and the Android e2e still grants (device reads `granted=true`), revokes (`false`), and resets, with 18 already-denied permissions left untouched and the camera revoked 868 ms after the click.

expo/serve-sim#144 stays open. iOS follows when serve-sim serves permissions over HTTP, or when serve-sim moves in-repo.

PR #93, opened 2026-09-10
feat(hub): show the Android foreground app's launcher icon

vonovak:
## Why

`ForegroundApp.iconDataUrl` has existed since serve-sim filled it, and the Current app
section renders it. Android never had a value to render, so `CurrentAppSection` hid the
whole identity row behind a platform check. serve-emu now extracts the launcher bitmap
from the installed APK, so an Android device shows the same name and icon row as iOS.

The extraction lives in a separate route rather than in `/api/foreground`, which the Hub
polls every 5 s. Resolving one icon costs 2 to 4 s of `adb pull` and `aapt2`, so the Hub
asks for it once per package and caches the answer.

## Scope

serve-emu gains `GET /api/apps/icon?packageName=`, wired through all three route chains
(`src/middleware.ts`, `src/server.ts`, `src/api/routes/applications.ts`) plus
`src/shared/api-contracts.ts` and its parser registry. The pipeline lives in the new
`src/apk-icon.ts`: `pm path` selects the base APK, `adb pull` copies it into a temp
directory, `aapt2 dump badging` names the icon entry, `aapt2 dump xmltree` reads the
adaptive icon's `foreground` drawable id, `aapt2 dump resources` resolves that id to the
densest bitmap file, and `unzip -p` extracts the bytes. Two tables do the picking: density
by qualifier, and MIME type by magic bytes. The response is
`{ok, packageName, icon: {mimeType, data} | null}`, with `data` as base64.

On the Hub, `hub-client/src/android-app-icon.ts` fetches and caches the data URL,
`useAndroidDevice` merges it into `foregroundApp` once per package, and
`CurrentAppSection` drops the `showIdentity` platform check.

Out of scope: no icon in `/api/foreground`, no APK disk cache, and no `aapt` v1 fallback.

## Tradeoffs

The route returns the adaptive icon's foreground layer alone, so the icon renders without
its background layer. Compositing the two layers would need image decoding on the serving
thread. `icon: null` is the honest answer for a vector-only launcher icon; a missing
`aapt2`, a failed pull, or an invalid package name is a 400 instead.

`aapt2 dump resources` prints 60,213,106 bytes for `com.android.settings`, well past the
executor's 8 MB output budget, so that one stage streams its stdout and kills the child at
the end of the target block instead of going through `execText`.

## Blast Radius

The serve-emu surface grows by one GET route; nothing existing changes behavior. The route
count assertions in `tests/api-routes.test.ts` and `tests/api-contracts.test.ts` move from
56/39 to 57/40. Every adb and aapt2 call runs in the `background` exec lane, so the icon
pipeline cannot starve interactive input.

On the Hub, the identity row now renders for every platform, which is a visible change for
Android users and the reason `inspectorSections.test.tsx` flips one assertion. A backend
without `aapt2` leaves the placeholder icon exactly as before.

## Verification

serve-emu: `bun run typecheck`, `bun test` (934 pass), and `bun run docs:check` all clean.
Hub: `bun run lint`, `bun run typecheck`, `bun run test` (hub-client 186, hub-components 24,
expo-device-hub 174, all passing), and `bunx tsc --noEmit` in `packages/expo-device-hub`
with vendor paths filtered prints nothing. `bunx changeset status` reports
`expo-device-hub` at minor. `/api/apps/icon` is present in the built
`vendor/serve-emu/dist/middleware.js`.

Against every launcher package on `emulator-5554`, `readAppIcon` resolves a bitmap for 19
of 22 and returns `null` for 3 with no errors. All three nulls are correct: Drive declares
`icon=''`, and Safety Hub and Voice Access resolve to `type=XML` vector drawables.

End to end through the Hub in headless Chrome, on the branch head. With Settings in the
foreground, the Current app section renders a 40 px `` whose `data:image/png;base64`
payload is 29,043 bytes, byte-identical to the bitmap a serve-emu-free shell script pulls
from the same APK with `adb pull` plus `aapt2` plus `unzip`. With Clock in the foreground,
the same row renders a 1,574 byte WebP, matching what the route returns for that package.
The shell readback script's simpler parser cannot follow Clock's nested `layer-list`
foreground, so the byte-for-byte independent comparison covers Settings and Chrome.

An adversarial review pass ran the pipeline against the live emulator. It found that
nested `<foreground>` children broke resolution for 5 packages, that a 400 response poisoned
the client cache for the tab's lifetime, and that concurrent requests held one full APK copy
each. All three are fixed in this branch and covered by unit tests or the device sweep above.


vonovak:
Root verifier verdict for this link: **PASS**, with one disclosed limit.

- Gates rerun in a fresh worktree at `979d9ec`: Hub `bun run lint`, `bun run typecheck`, `bun run test --force` exit 0; `tsc --noEmit` in `packages/expo-device-hub` shows no error outside `vendor/`; serve-emu `bun run typecheck` 0, `bun test` 0 fail, `docs:check` 0; the built `vendor/serve-emu/dist/middleware.js` serves `/api/apps/icon`; `changeset status` bumps `expo-device-hub` only.
- Receipts: `apk-icon.ts` is pure parsers over two tables (density by qualifier, MIME by magic bytes) around a pull, badging, unzip pipeline; the client caches one icon per base, device and package and evicts on failure; `CurrentAppSection` drops the iOS-only identity gate.
- Runtime, independent rerun from the verifier worktree on port 8104: with Settings in the foreground the Hub rendered a 40 px PNG of 29,043 bytes, byte-identical to the serve-emu-free `extract-icon.sh` readback of the APK. With Clock in the foreground the Hub rendered a 1,574-byte WebP; the independent extractor cannot follow Clock's nested layer-list, so that leg is proven only against the route's own output, as the PR body states.
- Rebased onto `vonovak/permissions` (link 6); the verdict carries over.

vonovak:
Restack note: rebasing onto link 6 conflicted in the serve-emu CHANGELOG, the three route chains, the contracts, the dependency type and the route tests (all add/add against link 6's permissions routes), plus one import block in `useAndroidDevice.ts` and the Android current-app fixture. The root kept both sides and added one commit that sets the stacked totals (63 method/path pairs across 45 paths), merges the two packageName query cases into one request builder, and lets the Android current-app test expect both the activity charts and the icon row. serve-emu tests (975) and the full Hub gates pass on the rebased head.

vonovak:
Rebased once more onto the corrected permissions tip (serve-sim pin restored to `a08fd7e`). No source change in this link.

krystofwoldrich-agent (commented):

**Verdict:** APPROVE
**Findings:** 0 critical · 0 design · 2 suggestion · 1 nit

The PR adds `GET /api/apps/icon` to serve-emu and wires it through all three route chains plus the contract map and parser registry, following the pattern the permissions routes use in #92. The pipeline in `src/apk-icon.ts` keeps every input bounded: `packageName` goes through the existing `PACKAGE_RE` validator, all subprocess calls pass argument arrays, the `unzip` output is capped at 4 MB, and the streamed `aapt2 dump resources` has its own timer. Extractions run one at a time and the temp directory is removed in `finally`. On the Hub, `getAndroidAppIcon` mirrors the existing `getIosAppDetails` cache shape, and `carryForwardAppIcon` covers the poll-overwrites-icon case with tests. No blocking concern; the items below are optional.

- [suggestion] `packages/@expo/hub-client/src/__tests__/android-app-icon.test.ts:62` — `rejects.toThrow('400')` is not awaited, so the assertion cannot fail the test.
- [suggestion] `packages/serve-emu/packages/serve-emu/src/apk-icon.ts:296` — `iconCache` has no eviction; a reinstall adds a new APK-path key and keeps the old entry for the process lifetime.
- [nit] `packages/@expo/hub-client/src/android-app-icon.ts:33` — client cache key omits the APK path, so a reinstalled app keeps its old icon until reload (matches the iOS cache behaviour).

Checks: serve-emu `bun test` on the three touched test files 43 pass; hub-client `android-app-icon.test.ts` 9 pass; `sync-readme.ts --check` synchronized; `tsc --noEmit` for serve-emu reports no errors in the changed files (remaining errors are missing `node-datachannel` types in the review worktree). hub-client `tsc` and `inspectorSections.test.tsx` could not run in the worktree (`react` not resolvable).

packages/@expo/hub-client/src/__tests__/android-app-icon.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  test('returns null when the package exposes no bitmap', async () => {
+    const { fetchImpl } = jsonFetch({ ok: true, packageName: 'com.example.app', icon: null });
+
+    expect(await fetchAndroidAppIcon(BASE, null, 'com.example.app', fetchImpl)).toBeNull();
+  });
+
+  test('throws on a failure response so the cache does not keep the miss', async () => {
+    const { fetchImpl } = jsonFetch({ ok: false, error: 'packageName is invalid' }, 400);
+
+    expect(fetchAndroidAppIcon(BASE, null, 'com.example.app', fetchImpl)).rejects.toThrow('400');
```
krystofwoldrich-agent: **[suggestion]** This `expect(...).rejects.toThrow('400')` is not awaited, so the test completes before the assertion settles. If `fetchAndroidAppIcon` stopped throwing on a 400, the test would still pass. The test at line 78 uses `await expect(...).rejects`; do the same here.

packages/serve-emu/packages/serve-emu/src/apk-icon.ts
```diff
@@ -0,0 +1,313 @@
...
+function queueExtraction(serial: string, baseApkPath: string): Promise<AppIcon | null> {
+  const next = extractions.then(
+    () => pullAndExtract(serial, baseApkPath),
+    () => pullAndExtract(serial, baseApkPath),
+  );
+  extractions = next.catch(() => {});
+  return next;
+}
+
+const iconCache = new Map<string, Promise<AppIcon | null>>();
```
krystofwoldrich-agent, line 296: **[suggestion]** `iconCache` is keyed by `serial:packageName:baseApkPath` and never evicts. A reinstall changes the APK path (`/data/app/~~hash==/pkg-xyz==/base.apk`), so it adds a new entry and the old one stays for the process lifetime. Each entry can hold up to ~5.3 MB of base64 under the 4 MB `MAX_ICON_BYTES` cap. Real icons are small (29 KB and 1.5 KB in the PR description), so the impact is low. Deleting the previous entry for the same `serial:packageName` when the APK path changes would keep the map bounded by installed packages.

packages/@expo/hub-client/src/android-app-icon.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const iconCache = new Map<string, Promise<string | null>>();
+
+export function getAndroidAppIcon(
+  baseUrl: string,
+  device: string | null,
+  packageName: string,
+  fetchImpl?: FetchImpl,
+): Promise<string | null> {
+  const key = `${baseUrl}:${device ?? ''}:${packageName}`;
```
krystofwoldrich-agent, line 45: **[nit]** The server cache key includes the installed APK path, so a reinstalled app returns its new icon. This key is `baseUrl:device:packageName`, so the Hub tab keeps the old icon until reload. This matches the iOS `detailsCache` in `ios-app-details.ts` ("immutable per installed build"), so it is consistent. Noting it because the server side went one step further and the two now differ.

vonovak:

vonovak:

**Act on**
- `android-app-icon.ts:16-22` casts the response instead of parsing it, then puts `mimeType` and `data` straight into an `` data URL. Every sibling module in the stack narrows from `unknown`, and serve-emu ships an unused `parseAppIconResponse`. Validate `ok`, `mimeType` against the four allowed values, and `data`. (interrogate reviewers A and B)
- `android-app-icon.test.ts:62`: `expect(...).rejects.toThrow('400')` is not awaited, so the test passes even if the function stops throwing. Await it like the test at line 78. (automated review; confirmed on the current head)
- `parseAppIconResponse` has no test, `server-request-gates.test.ts` lacks the `["POST", "/api/apps/icon"]` row, and the `ForegroundApp.iconDataUrl` doc still says "iOS only today". (interrogate B)
- Commit history: the union commit for the serve-emu route (`cf3d890` on the current head) carries a syntax error in `api-routes.test.ts` that only the tip reconciliation commit repairs. Fold the route-table hunk into that commit and the fixture hunk into the Hub commit with an autosquash rebase so each commit builds. (interrogate A)

**Consider**
- Two caches never evict. The server key includes the APK path, so a reinstall adds an entry and keeps the old one for the process lifetime. The client key has no APK identity, so a reinstalled app keeps its old icon for the tab, and a `null` result is cached forever, so a failed extraction never retries. The client cache matches the iOS `detailsCache` convention, and the effect already refetches once per app change, so dropping the client cache is a defensible simplification. (interrogate A and C, automated review)

**Noted**
- The Clock icon leg of the e2e is proven only against the route's own output; the independent extractor cannot follow a nested layer-list. Disclosed in the PR body. (root verdict)

vonovak:
Fix round on top of `50861cd`, new head `768b2a6`.

**Fixed**
- `9ee1202` awaits the `rejects.toThrow('400')` assertion at `android-app-icon.test.ts:62`. Worth noting for the next reader: Bun already fails that test when the function stops throwing, because it reports the unhandled rejection. I mutated `fetchAndroidAppIcon` to `return null` on a non-ok response and the un-awaited version failed too. The await is still correct, so it stays.
- `c4baa15` replaces the cast in `android-app-icon.ts` with `parseAndroidAppIcon`, which rejects a non-envelope payload, a non-object `icon`, a `mimeType` outside the four allowed values, and an empty or non-string `data`. Only an explicit `icon: null` still means "no bitmap", which matches serve-emu's `parseAppIconResponse`. Four new tests; the three rejecting ones fail against the old permissive body.
- `dff9733` exports `parseAppIconResponse` and covers it in `api-contracts-complete.test.ts` (all four MIME types, `icon: null`, plus rejections for `ok: false`, a missing `mimeType`, `image/svg+xml`, a non-string `data`, and a missing `packageName`), and adds `["POST", "/api/apps/icon"]` to `server-request-gates.test.ts`. Removing the route's method guard makes that row fail.
- `768b2a6` drops "iOS only today" from the `ForegroundApp.iconDataUrl` doc.

**Left**
- The commit-history fold is the root's call, not mine; folding `cf3d890` would rewrite commits the stack is already rebased onto.
- The two caches stay as they are (Consider, not Act on).
- The client does not import serve-emu's `APP_ICON_MIME_TYPES`. `@expo/hub-client` has no runtime dependencies, and adding an edge to a backend package for four strings is not worth it. The duplicate constant names its source of truth in a comment.

**Gates on `768b2a6`**: Hub `lint`, `typecheck`, `test --force` (274 hub-client, 181 Hub, 975 serve-emu, 58 hub-components, 121 android-utils, 44 apple-utils, 0 fail); `bunx tsc --noEmit` in `packages/expo-device-hub` shows only the three pre-existing `serve-emu-options.test.ts` errors from main; serve-emu `typecheck`, `bun test`, `docs:check` green; `build:vendor` rerun.

**E2E rerun** under the lane on this head: Settings renders a 29,043-byte PNG that is byte-identical to the APK's own launcher bitmap, Clock renders its 1,574-byte WebP. Both live MIME types pass the new allowlist. The run still exits 1 on the disclosed limit: the independent extractor cannot follow Clock's nested layer-list.

vonovak:
## Fix round and restack

All act-on items landed at `b2e8efc`: the icon response is parsed rather than cast, rejecting a bad envelope, a `mimeType` outside the four allowed values, and empty data; the failure assertion is awaited; `parseAppIconResponse` is exported and covered; the `POST /api/apps/icon` method gate has a row; and the stale `iconDataUrl` doc is fixed. The e2e still renders Settings as a 29,043-byte PNG byte-identical to the APK bitmap and Clock as its WebP, which proves the stricter parse did not reject the live types.

krystofwoldrich-agent (approved):
[nit] As a follow-up, we could look into combining the adaptive icon foreground and background layers to better match the visuals users see in the emulator launcher and app drawer.

krystofwoldrich (approved):

PR #94, opened 2026-09-10
Add strict host hardware H.264 encoding and probes

krystofwoldrich:
Adds strict host H.264 hardware encoding for Android gRPC screenshots while preserving the software FFmpeg command byte for byte. macOS uses VideoToolbox; Linux tries NVENC then VAAPI. A pinned backend is the only candidate, and hardware selection never falls back to software.

Each hardware backend must pass a real bounded 256×256 smoke encode with SPS/PPS, an IDR and multiple AUDs. Smoke processes use the shared ProcessExecutor with finite stdin, background lane admission, deadlines, output bounds and cancellation. Concurrent requests share a probe per configuration; cancelling one caller leaves other callers running, and cancelling every caller allows a clean retry. Successful probes are cached; failures remain retryable. A runtime-failure hook invalidates cached hardware success for the integration layer.

Backend arguments disable reordering/lookahead as appropriate and add AUDs. VideoToolbox requires `-allow_sw 0` and `-flags +low_delay`; the latter was verified necessary for delivery after one idle duplicate before stdin closes. The parser rejects excessive output without access-unit boundaries.

Validation: 63 focused executor/encoder tests, including real VideoToolbox output, subprocess accounting and stdin errors, concurrent cancellation/retry, runtime cache invalidation, the default 3000ms deadline message, and exact software argument regression. The combined stack passes all 973 serve-emu tests, critical coverage, typechecks, builds, documentation checks and package smoke; the latest Ubuntu CI check passes.

Live validation used Apple M4 Pro, macOS 26.6.2 and FFmpeg 9.0.1. Linux hardware execution remains unverified. Backend spike findings are recorded in the package documentation.

PR 1 of 3.

— Codex (GPT-6)

### DIFF
--- a/packages/serve-emu/packages/serve-emu/src/h264-encoder.ts
+++ b/packages/serve-emu/packages/serve-emu/src/h264-encoder.ts
@@ -2,6 +2,8 @@ import {
   spawn,
   type ChildProcessWithoutNullStreams,
 } from "node:child_process";
+import { access } from "node:fs/promises";
+import { constants } from "node:fs";
 import {
   execText,
   type ExecOpts,
@@ -12,17 +14,25 @@ import type { VideoFrame } from "./scrcpy.ts";
 /**
  * Host-side H.264 encoder used by the emulator gRPC screenshot source.
  *
- * RGB frames or complete PNG images enter through stdin and ffmpeg/libx264
- * writes Annex-B H.264 to stdout. `aud=1` gives the parser an explicit
- * access-unit boundary, while zerolatency and disabled B-frames keep input
+ * RGB frames or complete PNG images enter through stdin and ffmpeg writes
+ * Annex-B H.264 to stdout. Every backend inserts AUDs for explicit
+ * access-unit boundaries, while low-delay settings and disabled B-frames keep input
  * timestamps paired with output access units in submission order.
  */
+
+export type FfmpegEncoderName =
+  | "libx264"
+  | "h264_videotoolbox"
+  | "h264_nvenc"
+  | "h264_vaapi";
 
 export type H264EncoderInputFormat = "rgb24" | "png";
 
 export type H264EncoderOpts = {
   width: number;
   height: number;
+  /** Resolved ffmpeg backend. Defaults to the existing software encoder. */
+  encoderName?: FfmpegEncoderName;
   /** Input written to ffmpeg. Defaults to fixed-size raw RGB frames. */
   inputFormat?: H264EncoderInputFormat;
   /** Android display rotation quarter turns applied before encoding. */
@@ -43,6 +53,7 @@ const NAL_SPS = 7;
 const NAL_PPS = 8;
 const NAL_AUD = 9;
 const START_CODE = Buffer.from([0, 0, 0, 1]);
+const MAX_OUTPUT_WITHOUT_AUD_BYTES = 1024 * 1024;
 const MAX_PENDING_OUTPUT_BYTES = 64 * 1024 * 1024;
 const MAX_RAW_FRAME_BYTES = 512 * 1024 * 1024;
 const MAX_PNG_FRAME_BYTES = 64 * 1024 * 1024;
@@ -53,6 +64,8 @@ const PROCESS_GRACE_MS = 250;
 const PROCESS_TERM_MS = 500;
 const PROCESS_KILL_MS = 500;
 const FFMPEG_PROBE_TIMEOUT_MS = 10_000;
+const HARDWARE_PROBE_TIMEOUT_MS = 3_000;
+const HARDWARE_PROBE_MAX_BYTES = 1024 * 1024;
 const FFMPEG_PROBE_MAX_BYTES = 8 * 1024 * 1024;
 const MAX_FFMPEG_STDERR_BYTES = 16 * 1024;
 
@@ -131,6 +144,13 @@ function validateOptions(opts: H264EncoderOpts): number | null {
     throw new RangeError("quarterTurn must be an integer from 0 through 3");
   }
 
+  if (
+    opts.encoderName !== undefined &&
+    !["libx264", "h264_videotoolbox", "h264_nvenc", "h264_vaapi"].includes(opts.encoderName)
+  ) {
+    throw new RangeError(`unsupported ffmpeg H.264 encoder ${opts.encoderName}`);
+  }
+
   const inputFormat = opts.inputFormat ?? "rgb24";
   if (inputFormat !== "rgb24" && inputFormat !== "png") {
     throw new RangeError(`unsupported H.264 encoder input format ${inputFormat}`);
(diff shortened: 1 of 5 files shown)


packages/serve-emu/packages/serve-emu/src/h264-encoder.ts
```diff
@@ -427,6 +542,254 @@ export function createFfmpegAvailabilityProbe(
...
+                await checkVaapiDevice(device);
+              } catch {
+                throw new Error(
+                  `cannot access VAAPI device "${device}"; check the render group, driver and SERVE_EMU_VAAPI_DEVICE`,
+                );
+              }
+            }
+            if (signal?.aborted) throw probeAbortReason(signal);
+            const smoke = await runSmoke(binary, ffmpegEncoderArgs({
+              width: 128, height: 128, fps: 30, bitRate: 1_000_000,
```
krystofwoldrich: **[design] [P1] Probe NVENC with a supported frame size**

This 128×128 smoke frame is below NVENC's minimum H.264 width of 145 on Turing and newer GPUs, a [hardware restriction confirmed by NVIDIA](https://forums.developer.nvidia.com/t/minimum-width-in-turing-gpus/155566). A healthy RTX host therefore fails the capability smoke test even though its normal emulator frames are encodable; a pinned NVENC selection fails outright, and automatic selection unnecessarily falls through to VAAPI. Increase the probe to a compatible size such as 256×256 and update both the input buffer and the real-hardware test fixture. Add an assertion covering the NVENC-compatible dimensions so the mock smoke success cannot mask this regression.

— Codex (GPT-6)

krystofwoldrich (commented):

**Verdict: REQUEST_CHANGES** · PR 1/3 in the stack · 1 design finding.

The strict resolver, bounded process lifecycle and preservation of the software command fit the existing encoder architecture. The smoke frame is too small for NVENC on Turing and newer GPUs, so working NVIDIA hardware can be rejected before normal emulator capture starts. Use compatible dimensions and update the smoke buffer and real-hardware fixture together; see the inline finding.

Validation: all 39 encoder tests passed, including real VideoToolbox output before stdin closes. The additional 186 focused server/router, gRPC session, contract, CLI and UI tests passed across the stack. Linux GPU execution was not available; the NVENC finding is grounded in NVIDIA's documented hardware restriction.

— Codex (GPT-6)

packages/serve-emu/packages/serve-emu/src/h264-encoder.ts
krystofwoldrich: Fixed in 4d69fc82. The smoke probe and real hardware fixture now use 256×256 frames, with matching RGB input size. A regression asserts the actual NVENC smoke arguments and byte count. All 40 encoder tests pass, including the two-frame VideoToolbox test before EOF, and serve-emu typechecks pass. The dependent PRs were rebased with gh stack. Linux hardware execution remains unverified on this Mac.

— Codex (GPT-6)

krystofwoldrich (commented):

**Verdict: APPROVE** · Follow-up review of `4d69fc82` · The earlier P1 NVENC probe-dimension finding is resolved.

The production smoke encode and real-hardware fixture now use 256×256 RGB24 frames. The NVENC regression verifies the actual `-video_size 256x256` argument and that stdin contains four complete frames, covering the false-unavailable path identified in the original review.

I independently inspected the scoped fix and reran the encoder suite: **40 passed, 0 failed**, including real VideoToolbox delivery after one idle duplicate while stdin remained open. Linux GPU execution remains unverified. No new findings in this fix; this removes the dependency blocker noted in my reviews of #95 and #96.

— Codex (GPT-6)

packages/serve-emu/packages/serve-emu/src/h264-encoder.ts
```diff
@@ -427,6 +544,254 @@ export function createFfmpegAvailabilityProbe(
...
+      finish(null, "SIGKILL");
+    }, PROCESS_KILL_MS);
+  };
+  const onAbort = () => fail(probeAbortReason(options.signal!));
+  if (options.signal?.aborted) {
+    onAbort();
+    return;
+  }
+  try {
+    proc = spawn(binary, args, { stdio: ["pipe", "pipe", "pipe"] });
```
krystofwoldrich-agent: **[design]** serve-emu routes short-lived subprocesses through `ProcessExecutor`, which enforces lane concurrency, queue limits, and `/health` reporting. This runner calls `spawn` directly because `ExecSpawner` has no stdin path. The resolver still passes `lane: "background"` to the smoke runner (line 762), but nothing reads it, so a hardware probe never counts toward executor activity or lane limits.

The runner has its own timeout, output cap, and kill fallback, so it is bounded. But it is a second subprocess policy next to the existing one. Extending the executor with an optional stdin buffer would keep one policy and make the `lane` option real. If that is out of scope, drop the misleading `lane` option from the smoke calls.

packages/serve-emu/packages/serve-emu/src/h264-encoder.ts
```diff
@@ -427,6 +544,254 @@ export function createFfmpegAvailabilityProbe(
...
+            });
+            if (signal?.aborted) throw probeAbortReason(signal);
+            if (smoke.error || smoke.status !== 0) {
+              throw new Error(smoke.timedOut
+                ? `smoke encode timed out after ${HARDWARE_PROBE_TIMEOUT_MS}ms`
+                : smoke.stderr.trim().slice(-MAX_FFMPEG_STDERR_BYTES) ||
+                  smoke.error?.message || `exit ${smoke.status}`);
+            }
+            validateSmokeOutput(smoke.stdout);
+            successes.set(key, encoderName);
```
krystofwoldrich-agent: **[suggestion]** The resolver caches successes after the probe finishes but does not track an in-flight probe. Two devices switching to hardware at the same time, or a router retry loop, start two ffmpeg smoke processes for the same cache key. A pending-promise map keyed like `successes` would fold these into one.

packages/serve-emu/packages/serve-emu/src/h264-encoder.ts
```diff
@@ -427,6 +544,254 @@ export function createFfmpegAvailabilityProbe(
...
+  try {
+    proc = spawn(binary, args, { stdio: ["pipe", "pipe", "pipe"] });
+  } catch (error) {
+    fail(error instanceof Error ? error : new Error(String(error)));
+    return;
+  }
+  options.signal?.addEventListener("abort", onAbort, { once: true });
+  timer = setTimeout(() => {
+    timedOut = true;
+    fail(new Error(`smoke encode timed out after ${options.timeout}ms`));
```
krystofwoldrich-agent: **[nit]** The timer uses `options.timeout ?? HARDWARE_PROBE_TIMEOUT_MS`, but this message interpolates `options.timeout` alone, so it can read `undefinedms`. The resolver overwrites the message, so only direct callers see it.

packages/serve-emu/packages/serve-emu/src/h264-encoder.ts
```diff
@@ -131,6 +146,13 @@ function validateOptions(opts: H264EncoderOpts): number | null {
     throw new RangeError("quarterTurn must be an integer from 0 through 3");
   }
 
+  if (
+    opts.encoderName !== undefined &&
+    !["libx264", "h264_videotoolbox", "h264_nvenc", "h264_vaapi"].includes(opts.encoderName)
```
krystofwoldrich-agent: **[nit]** This inline array repeats the `FfmpegEncoderName` union. A `const FFMPEG_ENCODER_NAMES = [...] as const` used for both the type and this check keeps them in sync.

krystofwoldrich-agent (changes requested):

## Review: Add strict host hardware H.264 encoding and probes (PR 1/3)

**Verdict:** REQUEST_CHANGES
**Findings:** 0 critical · 1 design · 1 suggestion · 2 nit

Adds backend argument builders, a strict hardware probe resolver, and a 1 MB no-AUD guard in the parser. The software ffmpeg command is preserved byte for byte and pinned by an exact-array test. The parser guard is correct: after the first AUD, `#emitCompleteAccessUnits` keeps the last AUD in `#nals`, so the guard only fires when a backend never emits AUDs.

The main concern is that the smoke probe spawns ffmpeg outside the project's bounded subprocess executor. See inline comments for details.

- **[design]** Smoke encode bypasses `ProcessExecutor` (`src/h264-encoder.ts:603`); the `lane: "background"` option passed at line 762 is never read.
- **[suggestion]** Concurrent hardware requests run duplicate probes (`src/h264-encoder.ts:772`).
- **[nit]** Timeout message can read `undefinedms` (`src/h264-encoder.ts:611`).
- **[nit]** Encoder name list duplicated in validation (`src/h264-encoder.ts:151`).

Stack note: layer 2 (#95) turns a single probe failure into a permanent UI lock-out, which conflicts with this layer's "failed probes remain retryable" design. See the review on #95.

packages/serve-emu/packages/serve-emu/src/h264-encoder.ts
krystofwoldrich: Fixed in 77e67b0b. ProcessExecutor now accepts a finite stdin buffer; smoke encoding uses execBuffer with its real background lane, queue admission, deadlines, output limits, cancellation and activity reporting. The separate spawn/timeout policy is removed. Tests cover executor accounting, stdin delivery and EPIPE handling, alongside the real VideoToolbox encode.

— Codex (GPT-6)

packages/serve-emu/packages/serve-emu/src/h264-encoder.ts
krystofwoldrich: Fixed in 77e67b0b. Probes share an in-flight entry keyed by the same binary/platform/backend/device configuration. Each caller can cancel independently; only the last departing waiter cancels the shared process. Regressions cover concurrent callers, cancellation isolation, complete cancellation, overlapping retry cleanup and stale results.

— Codex (GPT-6)

packages/serve-emu/packages/serve-emu/src/h264-encoder.ts
krystofwoldrich: Fixed in 77e67b0b by using the shared executor's effective deadline. Direct smoke calls without a timeout now report 3000ms; a regression exercises that default.

— Codex (GPT-6)

packages/serve-emu/packages/serve-emu/src/h264-encoder.ts
krystofwoldrich: Fixed in 77e67b0b. FFMPEG_ENCODER_NAMES now defines both the encoder-name union and runtime validation.

— Codex (GPT-6)

PR #95, opened 2026-09-10
Expose hardware encoder selection throughout serve-emu

krystofwoldrich:
Threads `encoder: software|hardware` through Android gRPC sessions, CLI, router, standalone server, API contracts and the standalone selector. Software remains the default; scrcpy ignores the setting. Diagnostics and stream-mode responses report the resolved FFmpeg encoder name.

Source changes use staged replacement: a failed hardware attempt leaves the current capture and session generation intact. Hardware stays selectable after failure, with the last error retained as an advisory diagnostic and cleared after a successful probe. Unexpected runtime hardware encoder exits invalidate cached probe success; intentional close, replacement and abort do not. The diagnostics constructor now accepts named options.

Validation: 91 focused session/router/standalone tests, including fail-once/retry-success on both API implementations, runtime failure reporting, expected encoder lifecycle callbacks, and software isolation. The combined stack passes 973 serve-emu tests, critical coverage, typechecks, builds, documentation checks and package smoke. The latest Ubuntu CI check passes.

The built Hub was also tested in the Codex browser with WebRTC and a 60 FPS cap: one simulated hardware smoke failure preserved live Software at generation 0 and left Hardware enabled; retrying without a server restart produced live VideoToolbox video at generation 1 and cleared the error.

PR 2 of 3; depends on the encoder/probe layer in #94.

— Codex (GPT-6)

### DIFF
--- a/packages/serve-emu/packages/serve-emu/src/grpc-session.ts
+++ b/packages/serve-emu/packages/serve-emu/src/grpc-session.ts
@@ -25,6 +25,7 @@ import { execText, type ExecResult } from "./exec.ts";
 import {
   H264Encoder,
   resolveFfmpegEncoder,
+  reportFfmpegEncoderFailure,
   type FfmpegEncoderName,
   type H264EncoderInputFormat,
   type H264EncoderOpts,
@@ -209,12 +210,17 @@ export class GrpcCaptureDiagnosticsTracker {
   #acceptedEncoderWrites = 0;
   #encoderBackpressureRejections = 0;
 
-  constructor(
-    imageMode: GrpcImageMode,
+  constructor({
+    imageMode,
     windowCapacity = CAPTURE_DIAGNOSTIC_WINDOW,
     cadenceIdleResetMs = CAPTURE_CADENCE_IDLE_RESET_MS,
-    encoderName: FfmpegEncoderName = "libx264",
-  ) {
+    encoderName = "libx264",
+  }: {
+    imageMode: GrpcImageMode;
+    windowCapacity?: number;
+    cadenceIdleResetMs?: number;
+    encoderName?: FfmpegEncoderName;
+  }) {
     if (!Number.isFinite(cadenceIdleResetMs) || cadenceIdleResetMs <= 0) {
       throw new RangeError("cadence idle reset must be a positive number");
     }
@@ -1374,6 +1380,7 @@ export type GrpcSessionEncoder = {
 
 export type GrpcSessionRuntime = {
   resolveEncoder(encoder: GrpcEncoder, signal: AbortSignal): Promise<FfmpegEncoderName>;
+  reportEncoderFailure(encoderName: FfmpegEncoderName, message: string): void;
   ensureEndpoint(serial: string, signal: AbortSignal): Promise<GrpcEndpoint>;
   createClient(endpoint: GrpcEndpoint): GrpcSessionClient;
   createEncoder(options: H264EncoderOpts): GrpcSessionEncoder;
@@ -1644,6 +1651,7 @@ const defaultReadDisplaySizeSignal: NonNullable<
 
 const DEFAULT_GRPC_SESSION_RUNTIME: GrpcSessionRuntime = {
   resolveEncoder: resolveFfmpegEncoder,
+  reportEncoderFailure: reportFfmpegEncoderFailure,
   ensureEndpoint: ensureEmulatorGrpcEndpoint,
   createClient: (endpoint) => new EmulatorGrpcClient(endpoint),
   createEncoder: (options) => new H264Encoder(options),
@@ -1761,9 +1769,10 @@ export async function startGrpcSession(
     options.signal?.removeEventListener("abort", abortFromParent);
     throw error;
   }
-  const captureDiagnostics = new GrpcCaptureDiagnosticsTracker(
-    imageMode, undefined, undefined, encoderName,
-  );
+  const captureDiagnostics = new GrpcCaptureDiagnosticsTracker({
+    imageMode,
+    encoderName,
+  });
   const client = runtime.createClient(endpoint);
   const listeners = new Set<(failure: StreamFailure) => void>();
   const packetQueue = new GrpcVideoPacketQueue();
@@ -1934,6 +1943,7 @@ export async function startGrpcSession(
       sessionMeta.width = size.width;
       sessionMeta.height = size.height;
     }
+    let stopping = false;
     const next = runtime.createEncoder({
       encoderName,
       width: latest.width,
@@ -1947,7 +1957,13 @@ export async function startGrpcSession(
         if (!frame.isConfig) encoderHasOutput = true;
         pushPacket(frame);
       },
-      onExit: (message) => emitFatal({ message, code: "encoder-exit" }),
+      onExit: (message) => {
+        if (stopping || closed || lifetime.signal.aborted || fatalFailure) return;
+        if (encoderName !== "libx264") {
+          runtime.reportEncoderFailure(encoderName, message);
+        }
+        emitFatal({ message, code: "encoder-exit" });
+      },
     });
     if (restart.announceSize) {
       pushPacket({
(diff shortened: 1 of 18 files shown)


krystofwoldrich (commented):

**Verdict: APPROVE for this PR's diff** · PR 2/3 in the stack · No additional actionable findings.

Encoder selection follows the existing staged replacement path, survives settings/image-mode changes and device switching, and reports the resolved backend through the API and diagnostics. Failed replacement leaves the current capture and generation intact. The stack remains blocked by the NVENC smoke-frame finding in #94; fix that dependency before merging.

Validation: 186 focused tests passed across gRPC sessions, standalone server/router switching, the serve-emu selector, Hub contracts, CLI option forwarding and inspector UI. The separate encoder audit passed all 39 encoder tests.

— Codex (GPT-6)

packages/serve-emu/packages/serve-emu/src/middleware.ts
```diff
@@ -2509,6 +2557,10 @@ export function createRouter(
         defaults.grpcImageMode ??
         DEFAULT_GRPC_IMAGE_MODE,
     ),
+    encoder: encoderForApp(app, encoderOverrides.get(serial) ?? defaults.encoder ?? DEFAULT_GRPC_ENCODER),
+    encoderName: encoderNameForApp(app),
+    availableEncoders: hardwareEncoderError() ? ["software"] : [...GRPC_ENCODERS],
```
krystofwoldrich-agent: **[design]** `availableEncoders` drops `hardware` whenever `getHardwareEncoderError()` is set. The resolver only clears that error on a later *successful* hardware probe, and a hardware probe only runs when a caller requests `encoder: "hardware"`. Both UIs (the panel in this PR and the Hub sidebar in #96) disable the Hardware option when it is absent from this list, and the Hub hook refuses the call. So a transient failure, such as the 3 second smoke timeout on a loaded host, locks Hardware out of the UI until the server restarts. Only a raw `PUT /api/stream-mode` can recover.

This contradicts the layer 1 design note that "failed probes remain retryable" and its test of the same name. Keep `hardware` in `availableEncoders` and treat `hardwareEncoderError` as an advisory message, or clear the error after a time window. The same logic is duplicated in `server.ts:1713`.

packages/serve-emu/packages/serve-emu/src/server.ts
```diff
@@ -1673,11 +1700,18 @@ export async function startServer(
...
+    contextEncoders.get(context) ?? encoderForSerial(context.serial);
+
   const streamModeResponse = (context: DeviceContext) => ({
     ok: true as const,
     serial: context.serial,
     mode: context.stream.mode,
     grpcImageMode: grpcImageModeForContext(context),
+    encoder: encoderForContext(context),
+    encoderName: context.stream.diagnostics?.().grpcCapture?.encoderName ?? null,
+    availableEncoders: getHardwareEncoderError() ? ["software"] : [...GRPC_ENCODERS],
```
krystofwoldrich-agent: **[design]** Same one-way availability issue as `middleware.ts:2562`. See that comment.

packages/serve-emu/packages/serve-emu/src/grpc-session.ts
```diff
@@ -1922,6 +1935,7 @@ export async function startGrpcSession(
       sessionMeta.height = size.height;
     }
     const next = runtime.createEncoder({
+      encoderName,
```
krystofwoldrich-agent, line 1948: **[suggestion]** The probe encodes four 256×256 frames. Real sessions use the device size, which can exceed backend limits or exhaust NVENC session slots. If this ffmpeg exits, `onExit` raises an `encoder-exit` fatal (line 1950). Nothing in that path touches the resolver, so `successes` still holds the backend and `hardwareEncoderError` stays unset. The next session open skips the probe and reuses the same backend. Consider a `reportEncoderFailure(encoderName)` hook on the resolver that drops the cached success and records the message.

packages/serve-emu/packages/serve-emu/src/grpc-session.ts
```diff
@@ -1744,13 +1750,20 @@ export async function startGrpcSession(
...
+      options.encoder ?? DEFAULT_GRPC_ENCODER,
+      lifetime.signal,
+    );
     endpoint = await runtime.ensureEndpoint(serial, lifetime.signal);
   } catch (error) {
     options.signal?.removeEventListener("abort", abortFromParent);
     throw error;
   }
+  const captureDiagnostics = new GrpcCaptureDiagnosticsTracker(
+    imageMode, undefined, undefined, encoderName,
```
krystofwoldrich-agent: **[nit]** A fourth positional parameter after two optional ones forces `undefined, undefined` here. An options object would read better.

packages/serve-emu/packages/serve-emu/src/server.ts
```diff
@@ -1,3 +1,4 @@
+import { getHardwareEncoderError } from "./h264-encoder.ts";
```
krystofwoldrich-agent: **[nit]** This import sits ahead of the `node:` imports. Every other local import in this file is below them.

krystofwoldrich-agent (changes requested):

## Review: Expose hardware encoder selection throughout serve-emu (PR 2/3)

**Verdict:** REQUEST_CHANGES
**Findings:** 0 critical · 1 design · 1 suggestion · 2 nit

Threads an `encoder` setting through gRPC sessions, router, standalone server, contracts, and the standalone UI. The plumbing is consistent with the existing `grpcImageMode` handling, scrcpy strips the setting, and older clients parse cleanly through the `software` defaults.

The main concern is the availability model. Layer 1 (#94) makes probe failures retryable. This layer drops `hardware` from `availableEncoders` after one failure, and both UIs disable the option based on that list. The result is a one-way door until the server restarts. See inline comments.

- **[design]** One failed probe disables Hardware for the process lifetime (`src/middleware.ts:2562`, `src/server.ts:1713`).
- **[suggestion]** A cached hardware success is never invalidated by a runtime encoder failure (`src/grpc-session.ts:1938`).
- **[nit]** Positional diagnostics constructor with `undefined, undefined` (`src/grpc-session.ts:1765`).
- **[nit]** New import placed before Node built-ins (`src/server.ts:1`).

packages/serve-emu/packages/serve-emu/src/middleware.ts
krystofwoldrich: Fixed in a2228ea1. Hardware remains in availableEncoders; hardwareEncoderError is advisory and clears after a successful probe. Router and standalone tests cover failure followed by success while preserving the original stream/generation on failure. Verified in the Codex browser: one simulated probe failure kept Software live at generation 0, then a UI retry without server restart resumed VideoToolbox at generation 1.

— Codex (GPT-6)

packages/serve-emu/packages/serve-emu/src/server.ts
krystofwoldrich: Fixed in a2228ea1 in the standalone implementation too. Hardware stays selectable, and a dedicated standalone regression exercises a failed switch followed by successful retry.

— Codex (GPT-6)

packages/serve-emu/packages/serve-emu/src/grpc-session.ts
krystofwoldrich: Fixed across 77e67b0b and a2228ea1. Unexpected hardware encoder exits report through reportFfmpegEncoderFailure, invalidate cached success and retain the failure diagnostic. Intentional close/replacement, abort and software failures do not invalidate hardware. Tests cover those lifecycle distinctions plus a fresh probe after failure and stale-result protection.

— Codex (GPT-6)

packages/serve-emu/packages/serve-emu/src/grpc-session.ts
krystofwoldrich: Fixed in a2228ea1. GrpcCaptureDiagnosticsTracker now accepts named options for imageMode, encoderName and the optional timing settings; all call sites were updated.

— Codex (GPT-6)

packages/serve-emu/packages/serve-emu/src/server.ts
krystofwoldrich: Fixed in a2228ea1. The encoder import now follows the Node built-ins with the other local imports.

— Codex (GPT-6)

PR #100, opened 2026-09-10
Preserve right sidebar sections and scroll position

krystofwoldrich-agent:
Closing the right inspector previously discarded its expanded sections and scroll position. Keep a single inspector across docked and overlay layouts, and preserve its React state and DOM with an Activity boundary after the exit animation. Hidden inspector subscriptions are still cleaned up and its controls are inert. Logs and Events also retain their nested scroll positions when Activity reactivates effects with unchanged data.

Validation:
- Added a browser regression against the exported dashboard with an empty device host. It failed on upstream and now passes for normal and reduced motion, expanded/collapsed sections, scrolling, overlay dismissal, and resizing while open or closed.
- Added a populated inspector fixture covering both Logs and Events scroll positions on reopen; new log data follows the tail when the viewer is at the bottom, while reading older entries preserves the scroll position.
- Web export succeeds; all 177 expo-device-hub tests and 24 hub-components tests pass. Hub-components lint and type checking pass.
- `bunx tsc --noEmit --types bun` reports the same three existing errors in `serve-emu-options.test.ts` on this branch and upstream `main`.
- Recorded a 20-second verification on an actual streaming iPhone 17 Pro simulator (iOS 26.5): expanded settings, 24 real touch events, and a moving timer in Safari. The sidebar offset remains 650 → 650 px; the Events pane remains 150 → 150 px across closing/reopening.
- The live recording exposed replayed event snapshots resetting the scroll. The follow-up fix preserves history positions across replaced buffers and incoming entries, with browser regression coverage.
- Independent deep review posted one finding about populated panes; fixed in the follow-up commit with regression coverage.

To reproduce the browser check: build the workspace dependencies and web client, install Chromium with `bunx playwright install chromium`, then run `bun run test:sidebar` from `packages/expo-device-hub`. Set `SIDEBAR_VIDEO_DIR` to record the run.

### DIFF
--- a/packages/expo-device-hub/src/dashboard/RightSidebar.tsx
+++ b/packages/expo-device-hub/src/dashboard/RightSidebar.tsx
@@ -1,0 +1,91 @@
+import { bg, border } from '@expo/hub-components';
+import { Activity, type ReactNode } from 'react';
+
+import {
+  SIDEBAR_TRANSITION_EASING,
+  SIDEBAR_TRANSITION_MS,
+  useSidebarPresence,
+} from './useSidebarPresence';
+
+/** One inspector tree across closing, reopening, and docked/overlay layout changes. */
+export function RightSidebar({
+  open,
+  overlay,
+  width,
+  resizing,
+  topmost,
+  onDismiss,
+  children,
+}: {
+  open: boolean;
+  /** Whether the available space requires an overlay, even while closed. */
+  overlay: boolean;
+  width: number;
+  resizing: boolean;
+  topmost: boolean;
+  onDismiss: () => void;
+  children: ReactNode;
+}) {
+  const { present, reducedMotion, visible } = useSidebarPresence(open, true);
+  const backdropZIndex = topmost ? 12 : 10;
+  const panelTransition = reducedMotion
+    ? undefined
+    : `transform ${SIDEBAR_TRANSITION_MS}ms ${SIDEBAR_TRANSITION_EASING}`;
+
+  return (
+    <>
+      {overlay && present && (
+        <div
+          aria-hidden="true"
+          data-sidebar-backdrop="right"
+          onClick={onDismiss}
+          style={{
+            position: 'absolute',
+            inset: 0,
+            backgroundColor: bg.overlay,
+            opacity: visible ? 0.35 : 0,
+            pointerEvents: visible ? undefined : 'none',
+            transition: reducedMotion ? undefined : `opacity ${SIDEBAR_TRANSITION_MS}ms ease`,
+            zIndex: backdropZIndex,
+          }}
+        />
+      )}
+      <div
+        aria-hidden={!visible || undefined}
+        inert={!visible || undefined}
+        data-sidebar-docked={overlay ? undefined : 'right'}
+        data-sidebar-overlay={overlay ? 'right' : undefined}
+        data-state={visible ? 'open' : 'closed'}
+        style={{
+          position: overlay ? 'absolute' : undefined,
+          top: overlay ? 0 : undefined,
+          right: overlay ? 0 : undefined,
+          zIndex: overlay ? backdropZIndex + 1 : undefined,
+          width: overlay ? `min(${width}px, 100vw)` : visible ? width : 0,
+          height: '100%',
+          flexShrink: 0,
+          overflow: 'hidden',
+          pointerEvents: visible ? undefined : 'none',
+          transition:
+            overlay || reducedMotion || resizing
+              ? undefined
+              : `width ${SIDEBAR_TRANSITION_MS}ms ${SIDEBAR_TRANSITION_EASING}`,
+        }}>
+        <div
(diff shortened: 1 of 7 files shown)


packages/expo-device-hub/src/dashboard/RightSidebar.tsx
```diff
@@ -0,0 +1,91 @@
...
+            height: '100%',
+            backgroundColor: bg.default,
+            borderLeft: overlay ? `1px solid ${border.default}` : undefined,
+            transform: visible ? 'translateX(0)' : 'translateX(100%)',
+            transition: panelTransition,
+            willChange: 'transform',
+          }}>
+          {/* Preserve section state and DOM scroll offsets, but release subscriptions
+              after the exit animation while the inspector is hidden. */}
+          <Activity mode={present ? 'visible' : 'hidden'}>{children}</Activity>
```
krystofwoldrich-agent, line 86: **[design] Preserve the log/event pane scroll offset when Activity resumes.** Switching this boundary back to `visible` re-runs its descendants' effects. `packages/@expo/hub-components/src/dashboard/LogList.tsx` has an effect that unconditionally assigns `scrollTop = scrollHeight`, so preserving the DOM here does not preserve the nested Logs scroll position. In a browser fixture using the actual RightSidebar and LogSidebar with 100 unchanged log entries, I scrolled the log pane to 100, closed the sidebar past the exit animation, and reopened it: the offset became 3319 (the bottom), under both normal and reduced motion. No new entries arrived. Events uses the same LogList and is affected by the same lifecycle. Preserve the existing offset during reactivation and add a populated log/event regression; the current empty-device fixture only checks the outer `aside > div` scroller.

krystofwoldrich-agent (commented):

**Verdict: REQUEST_CHANGES**
**Findings:** 0 critical · 1 design · 0 suggestion · 0 nit

The single inspector tree preserves section expansion and the outer sidebar scroll position across closing and responsive layout changes. One scroll-preservation gap remains: reactivating the Activity boundary re-runs the existing LogList effect and sends a populated Logs pane to the bottom, even when no entries changed. The same component is used by Events. See the inline finding.

Validation: all 177 expo-device-hub tests passed; `bun run test:sidebar` passed for normal and reduced motion. A separate browser fixture using the actual RightSidebar and LogSidebar with 100 static log entries reproduced a scroll jump from 100 to 3319 after close/reopen in both motion modes. No product code was changed.

Submitted as a comment because the authenticated account is also the PR author; the review verdict above is REQUEST_CHANGES.

krystofwoldrich-agent (commented):

**Iteration 2 — Verdict: APPROVE**
**Outstanding findings: 0.**

Re-reviewed the follow-up diff at `8721bd8`. The original Logs/Events scroll-preservation finding is resolved: LogList now skips tail scrolling when Activity reactivates its effect with the same buffer, while changed buffers retain the existing tail-follow behavior. No new findings in this follow-up diff.

Independent verification: the original populated-log reproduction now retains offset 100 after close/reopen in both normal and reduced motion (previously jumped to 3319). The expanded `bun run test:sidebar` passes the dashboard and populated Logs/Events cases in both motion modes, including tail-follow after appending a log. Package tests: 177/177 expo-device-hub and 24/24 hub-components passed.

Submitted as COMMENT because the authenticated account is the PR author; the review verdict is APPROVE.

packages/expo-device-hub/src/dashboard/RightSidebar.tsx
krystofwoldrich-agent: Verified resolved at `8721bd8b860d1a12d9ef29dee54149d8f8192256` (automated re-review). I re-ran my original fixture against the fix: offset 100 remains 100 after close/reopen under both normal and reduced motion, rather than jumping to 3319. The new populated Logs/Events regression passes in both motion modes and confirms new log data still follows the tail. No further issue found in the follow-up diff.

krystofwoldrich-agent (commented):

**Iteration 3 — Verdict: APPROVE**
**Findings: 0.**

Reviewed only follow-up commit `4b7280f`. Tracking whether the viewer is at the bottom fixes the reconnect case where unchanged Events arrive in a fresh array, while preserving history positions during incoming data. The hidden-layout guard prevents Activity's zero-size layout from incorrectly enabling tail-follow. No new findings.

Verification: `bun run test:sidebar` passes the dashboard and populated Logs/Events regressions in both motion modes; all 24 hub-components tests pass. I also independently checked buffer replacement plus new data while hidden, reopening while following the tail, and 30 rapid appends: history offsets remain unchanged and tail-follow continues correctly under normal and reduced motion. The original close/reopen reproduction still retains offset 100.

Submitted as COMMENT because the authenticated account is the PR author; the review verdict is APPROVE.

krystofwoldrich:
https://github.com/user-attachments/assets/97589315-251b-4755-a550-9c9493ceb26f

PR #101, opened 2026-09-10
Default Hub video to 60 FPS

krystofwoldrich-agent:
Starting the Hub without an FPS override previously allowed Android middleware to fall back to 30 FPS. Default video to 60 FPS in both the standalone CLI and Expo DevTools plugin. The CLI forwards this default to Android and iOS, and explicit integer overrides from 1 to 120 continue to work.

Update CLI help and the README with the default and an overview of FPS and related performance flags. Include a patch changeset. Keep the optional CLI option type for hand-built options and the help path, with a comment explaining that choice.

Validation: 33 targeted tests passed across CLI parsing and Android/iOS option mapping, including the plugin fallback; git diff --check passed. Live device streaming was not tested.


packages/expo-device-hub/src/server/cli/options.ts
```diff
@@ -11,6 +11,7 @@ import {
 } from '../../transport';
 
 export const DEFAULT_PORT = 3400;
+export const DEFAULT_VIDEO_FPS = 60;
```
krystofwoldrich-agent, line 14: **[suggestion]** This default only reaches the standalone CLI. When the Hub runs as an Expo DevTools plugin, `SERVE_EMU_OPTIONS_ENV` is unset, `readStandaloneServeEmuOptions` returns `defaultServeEmuOptions()` without `maxFps`, and the serve-emu middleware falls back to 30 FPS (`packages/serve-emu/packages/serve-emu/src/middleware.ts:362`, `h264Fps: opts.maxFps ?? 30`). Adding `maxFps: DEFAULT_VIDEO_FPS` to `defaultServeEmuOptions()` in `serve-emu-options.ts`, or fixing the serve-emu fallback to use `SCRCPY_DEFAULTS.maxFps` (60), would give both paths the same default. Out of the PR's stated scope, so not blocking.

packages/expo-device-hub/src/server/cli/options.ts
```diff
@@ -214,7 +215,8 @@ export function parseCliOptions(args: string[]): CliOptions {
     50_000_000,
     true
   );
-  const videoFps = parseNumberOption(values['video-fps'], '--video-fps', 1, 120, true);
+  const videoFps =
+    parseNumberOption(values['video-fps'], '--video-fps', 1, 120, true) ?? DEFAULT_VIDEO_FPS;
```
krystofwoldrich-agent, line 220: **[nit]** `videoFps` is always a number after this line, but `CliOptions.videoFps` stays `number | undefined`. The `options.videoFps !== undefined` guards in `serve-emu-options.ts:71` and `serve-sim-options.ts:17` can no longer be false for parsed options. Either make the field required or keep the guards for hand-built `CliOptions`. Take it or leave it.

krystofwoldrich-agent (commented):

**Verdict:** APPROVE
**Findings:** 0 critical · 0 design · 1 suggestion · 1 nit

The PR defaults `videoFps` to 60 in `parseCliOptions` and documents the flag. The motivation checks out: the in-tree serve-emu middleware falls back to 30 FPS when `maxFps` is absent (`packages/serve-emu/packages/serve-emu/src/middleware.ts:362`), while serve-emu's own `SCRCPY_DEFAULTS.maxFps` and CLI default are 60. On iOS the change is a value no-op: serve-sim already defaults `h264Fps` to 60, and `{ transport: 'http', h264Fps: 60 }` normalizes to the same control settings as `undefined`. README claims (serve-sim `--mjpeg-fps` 1–120, `--video-fps` up to 140; serve-emu `--max-fps`, `--repeat-frame-ms`, `--key-frame-interval`) match the pinned sources. Tests: 33 pass in the three affected files. TypeScript check did not run (no installed dependencies in the review worktree).

- **[suggestion]** `packages/expo-device-hub/src/server/cli/options.ts:14` — The DevTools plugin path still gets 30 FPS on Android because `defaultServeEmuOptions()` sends no `maxFps`. Out of the stated scope; not blocking.
- **[nit]** `packages/expo-device-hub/src/server/cli/options.ts:219` — `videoFps` is always set now, but the type stays optional and the `!== undefined` guards in the option mappers can no longer be false.

PR #104, opened 2026-09-10
Forward gRPC frames as soon as FFmpeg is ready

krystofwoldrich:
gRPC frames currently wait for a local FPS pacer and retry rejected FFmpeg writes on an 8 ms timer. Submit usable frames immediately when FFmpeg is ready and continue from its `drain` event, retaining the newest pending image. This is the only encoder handoff path, with no experimental switches.

Keep consuming RGB888 screenshots during encoder backpressure so HTTP/2 does not become a FIFO of stale frames. Receive work yields after each 64 KiB batch so a busy source cannot starve pipe callbacks; stream and connection receive windows are 8 MiB. Track stdin's write result explicitly because Bun can report backpressure while `writableNeedDrain` remains false.

A one-shot five-second watchdog recovers a missed drain only when the encoder reports writable; otherwise it emits `encoder-exit`. Screenshot inactivity probes remain active. Guard synchronous construction callbacks and callbacks from replaced encoders. Continue startup boundary repeats until encoded output arrives, including on static RGB screens, then use the existing idle cadence. PNG predecode pacing and MMAP shared-memory selection remain in place. For RGB888, `--max-fps` configures the nominal encoder rate and idle cadence without capping fresh submissions; the documentation and CLI help explain this.

Validation:

- `bun run --filter serve-emu check`: 987 tests pass, critical coverage, typechecks, build and package smoke pass. Regressions cover an actual HTTP/2 full-size frame burst during encoder backpressure, lost drains, stall cancellation, startup failures and synchronous/stale callbacks.
- Hub: 184 tests pass; repository lint and Hub vendor/server builds pass.
- Live 540×1170 Android emulator: Bun/software and Node/VideoToolbox each delivered 10 frames. Node/VideoToolbox also passed on a static home screen. The rebuilt Node CLI streams the animated workload over WebRTC without experimental flags.
- Backpressure test with a 240 FPS synthetic source and FFmpeg limited to 30 FPS: P95 source-to-submit frame age fell from 170.12 ms to 41.84 ms (mean 90.97 → 12.56 ms; P90 154.26 → 27.88 ms), with unchanged 30 FPS output. Baseline is the earlier PR commit `43e447316`; each run measured 10 seconds after warm-up.
- Local production gRPC → FFmpeg rawvideo copy path reached 288.88 fresh FPS over 8 seconds. With a target 240 FPS source, it wrote 228.40 fresh FPS over 10 seconds; complete-image receipt to pipe-write completion averaged 3.51 ms, P90 7.17 ms, P95 9.73 ms. All completed writes matched FFmpeg's frame count, with zero write errors.
- The production software H.264 encoder accepted 237.46 fresh FPS over 10 seconds with the unrestricted synthetic source. These are short local capacity measurements at 540×1170, not emulator/browser display FPS. Older pending images are intentionally replaced during stalls, and synthetic frames compress easily. Benchmark artifacts remain local.

— Codex (GPT-6)

### DIFF
--- a/packages/serve-emu/packages/serve-emu/src/grpc-session.ts
+++ b/packages/serve-emu/packages/serve-emu/src/grpc-session.ts
@@ -17,6 +17,7 @@ import {
   type GrpcMessagePacingEvent,
   type GrpcMessagePacingDetail,
   type GrpcScreenshotImageSource,
+  type GrpcScreenshotReadControl,
   type ImageFormatRequest,
   type KeyboardEventRequest,
   type TouchPoint,
@@ -78,7 +79,6 @@ const ACCESS_UNIT_CADENCE_WINDOW = 8;
 const ACCESS_UNIT_CADENCE_OUTLIER_FLOOR_MS = 250;
 const ACCESS_UNIT_CADENCE_OUTLIER_MULTIPLIER = 4;
 const ACCESS_UNIT_SLOW_CADENCE_SIMILARITY = 1.5;
-const ENCODER_WRITE_RETRY_DELAY_MS = 8;
 const DEFAULT_IDLE_REPEAT_MS = 500;
 const FIRST_FRAME_TIMEOUT_MS = 10_000;
 const MAX_QUEUED_PACKET_BYTES = 64 * 1024 * 1024;
@@ -706,10 +706,8 @@ export function grpcImageModeBehavior(
   }
   return {
     encoderInputFormat: "rgb24",
-    // Drain raw RGB responses independently of encoder FPS. Pausing HTTP/2
-    // queues old pixels upstream; onImage keeps only the newest frame and the
-    // encoder write pacer decides when to submit it. MMAP also drains metadata
-    // immediately, then selects which notifications trigger memory reads.
+    // In-band RGB uses encoder readiness for flow control. MMAP drains
+    // metadata immediately, then selects notifications for memory reads.
     predecodeMaxFps: undefined,
     needsEncoderFollowUp: (repeat) => !repeat,
   };
@@ -1010,34 +1008,6 @@ export function resolveGrpcDisplayGeometry(options: {
       };
     },
   };
-}
-
-export class GrpcFrameWritePacer {
-  readonly #frameIntervalMs: number;
-  #nextFreshWriteAt = 0;
-
-  constructor(frameIntervalMs: number) {
-    if (!Number.isFinite(frameIntervalMs) || frameIntervalMs <= 0) {
-      throw new RangeError("frameIntervalMs must be a positive number");
-    }
-    this.#frameIntervalMs = frameIntervalMs;
-  }
-
-  reset(now: number): void {
-    this.#nextFreshWriteAt = now;
-  }
-
-  recordWrite(now: number, repeat: boolean, accepted = true): void {
-    if (repeat || !accepted) return;
-    this.#nextFreshWriteAt = Math.max(
-      this.#nextFreshWriteAt + this.#frameIntervalMs,
-      now + this.#frameIntervalMs,
-    );
-  }
-
-  waitMs(now: number): number {
-    return Math.max(0, this.#nextFreshWriteAt - now);
-  }
 }
 
 export class GrpcAccessUnitBoundaryCadence {
@@ -1357,6 +1327,7 @@ export type GrpcSessionClient = {
     signal: AbortSignal,
     options?: {
       maxFps?: number;
+      onReadControl?: (control: GrpcScreenshotReadControl) => void;
       onPacingEvent?: (
         event: GrpcMessagePacingEvent,
         detail: GrpcMessagePacingDetail,
@@ -1371,6 +1342,7 @@ export type GrpcSessionClient = {
 };
 
 export type GrpcSessionEncoder = {
+  readonly writable: boolean;
   readonly width: number;
   readonly height: number;
   readonly quarterTurn: QuarterTurn;
@@ -1739,7 +1711,6 @@ export async function startGrpcSession(
       ? configuredRepeatFrameMs
       : DEFAULT_IDLE_REPEAT_MS;
   const frameIntervalMs = 1_000 / maxFps;
-  const frameWritePacer = new GrpcFrameWritePacer(frameIntervalMs);
   const accessUnitBoundaryCadence = new GrpcAccessUnitBoundaryCadence(
     frameIntervalMs,
   );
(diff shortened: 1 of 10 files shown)


packages/serve-emu/packages/serve-emu/src/grpc-session.ts
```diff
@@ -1964,6 +1908,31 @@ export async function startGrpcSession(
...
+          stopping || closed || lifetime.signal.aborted || fatalFailure ||
+          encoderLifecycle?.current !== managed
+        ) return;
+        waitingForEncoder = false;
+        const repeat = latest === lastSubmittedImage;
+        if (!repeat) {
+          pendingRepeat = false;
+          writeFrame(false);
+        } else {
+          screenshotReadControl?.resume();
```
krystofwoldrich-agent: **[design]** After `resume()` the buffered HTTP/2 data is emitted in order, and the first message triggers `writeFrame(false)`, which pauses again once the frame write returns `false` (any full-frame write exceeds the pipe's 16 KiB highWaterMark). Under sustained backpressure this encodes the oldest buffered frame on every drain, with up to a full 8 MiB window of stale frames queued (about four frames at 540×1170). The README says only the newest pending image is retained; that no longer holds for RGB888 here. One option: defer the fresh write with `setImmediate` after `resume()` so buffered messages are parsed and `latest` is the newest before writing. Another: size the window to about one frame, or keep draining and use `drain` only to replace the retry timer.

packages/serve-emu/packages/serve-emu/src/grpc-session.ts
```diff
@@ -1856,66 +1828,36 @@ export async function startGrpcSession(
     ) {
       return;
     }
+    if (waitingForEncoder) {
```
krystofwoldrich-agent, line 1880: **[design]** This early return makes `drain` the only path out of `waitingForEncoder`. The idle timer only sets `pendingRepeat`, and `pause()` in `emulator-grpc.ts` disables the inactivity probe. If `drain` never fires (the PR notes Bun's `writableNeedDrain` is unreliable), the stream freezes with no fatal error or reconnect. Consider a bounded fallback: if `waitingForEncoder` persists beyond a few seconds, re-check `encoder.writable` and retry, or emit a fatal `encoder-exit`.

packages/serve-emu/packages/serve-emu/src/grpc-session.ts
```diff
@@ -1964,6 +1908,31 @@ export async function startGrpcSession(
         }
         emitFatal({ message, code: "encoder-exit" });
       },
+      onWritable: () => {
+        if (
+          stopping || closed || lifetime.signal.aborted || fatalFailure ||
+          encoderLifecycle?.current !== managed
+        ) return;
```
krystofwoldrich-agent: **[nit]** `managed` is a `const` declared after `runtime.createEncoder` returns. Any encoder that calls `onWritable` synchronously during construction would hit a TDZ ReferenceError here. Declaring `managed` before creating the encoder, or comparing against a local flag, removes the ordering dependency.

krystofwoldrich-agent (changes requested):

## Review: Forward gRPC frames as soon as FFmpeg is ready

**Verdict:** REQUEST_CHANGES
**Findings:** 0 critical · 2 design · 0 suggestion · 1 nit

The change replaces the local FPS pacer and the 8 ms retry timer with a drain-driven handoff. The `H264Encoder.writable` getter, the `onWritable` callback, and the encoder-replacement guards are clean, and the new integration tests cover the drain, stale-encoder, and pause/resume paths well. The main concern is what pausing the HTTP/2 stream does to frame freshness for RGB888 once ffmpeg is the bottleneck. A second concern is that the drain event is now the only recovery path, with the inactivity watchdog disabled while paused.

### Findings

#### [design] RGB888 encodes the oldest buffered frame after each drain, not the newest
`packages/serve-emu/packages/serve-emu/src/grpc-session.ts:1922`

On Node, a full-frame write to the ffmpeg pipe almost always returns `false` (the frame exceeds the 16 KiB highWaterMark), so `waitingForEncoder` becomes true after nearly every accepted write and `screenshotReadControl.pause()` runs. While paused, the emulator keeps sending until the 8 MiB stream window is full, and Node buffers that data in order. On `drain`, `onWritable` sees `latest === lastSubmittedImage`, calls `resume()`, and the first buffered message reaches `onImage` → `writeFrame(false)` → pause again. The result is a FIFO of window-size depth: at 540×1170 (about 1.9 MiB per frame) that is roughly four stale frames whenever the source outpaces the encoder. The previous design kept draining and only encoded the newest image; the README still says the newest pending image is retained, which no longer holds for RGB888 under sustained backpressure. The 239 FPS benchmark used a rawvideo copy, so the consumer never fell behind and this path was not measured.

Options: (a) after `resume()`, defer the fresh write with `setImmediate` so the buffered messages are parsed and `latest` is the newest before writing; (b) keep the pause but size the receive window at about one frame; (c) keep draining as before and only use `drain` to replace the retry timer. Either way, an end-to-end latency measurement with libx264 saturated would settle this.

#### [design] A missed `drain` now stalls the stream silently
`packages/serve-emu/packages/serve-emu/src/grpc-session.ts:1831`

The 8 ms retry loop was the only recovery when ffmpeg did not signal readiness. Now `writeFrame` returns early while `waitingForEncoder`, the idle timer only sets `pendingRepeat`, and `pause()` in `emulator-grpc.ts` clears the inactivity timer and bumps `activityGeneration`. If `drain` never fires (the PR itself notes that Bun reports backpressure inconsistently via `writableNeedDrain`), the session freezes with no fatal error, no probe, and no reconnect. A bounded fallback would keep the failure visible: for example, if `waitingForEncoder` persists for longer than a few seconds, re-check `encoder.writable` and either retry or emit a fatal `encoder-exit` so the client can reconnect.

#### [nit] `managed` is referenced before its declaration
`packages/serve-emu/packages/serve-emu/src/grpc-session.ts:1915`

`onWritable` closes over `managed`, which is declared with `const` after `runtime.createEncoder` returns. This is fine as long as no encoder implementation invokes `onWritable` synchronously during construction; a fake or a future backend that does would throw a ReferenceError from the TDZ. Declaring `managed` first (or capturing `next` and comparing via a local flag) removes the ordering dependency.

packages/serve-emu/packages/serve-emu/src/grpc-session.ts
krystofwoldrich: Fixed in c48cbc1da. RGB888 now keeps consuming gRPC frames independently of FFmpeg backpressure, replacing the pending image so the next drain submits the newest complete frame. Receive work yields after 64 KiB per event-loop turn to keep a busy source from starving pipe callbacks. A real HTTP/2 regression test sends eight full-size frames while the encoder is blocked and verifies that the next drain submits frame eight.

— Codex (GPT-6)

packages/serve-emu/packages/serve-emu/src/grpc-session.ts
krystofwoldrich: Fixed in c48cbc1da with a one-shot five-second watchdog when the encoder becomes backpressured. At the deadline, it resumes only if encoder.writable is true; otherwise it emits a fatal encoder-exit error. Incoming frames do not extend the deadline, and screenshot inactivity probes remain active. Drain, encoder replacement, and session close clear the watchdog. Regression tests cover missed-notification recovery, persistent stalls, startup failure, and cancellation.

— Codex (GPT-6)

packages/serve-emu/packages/serve-emu/src/grpc-session.ts
krystofwoldrich: Fixed in c48cbc1da by declaring managed as a nullable variable before runtime.createEncoder and guarding onWritable until it has been assigned. Synchronous construction callbacks are safely ignored, while callbacks from replaced encoders still fail the current-encoder identity check. Added a regression test whose encoder factory calls onWritable synchronously before returning; session startup succeeds without a TDZ error.

— Codex (GPT-6)

PR #105, opened 2026-09-10
ci: build expo-device-hub releases with EAS Workflows

krystofwoldrich:
Move expo-device-hub release builds from GitHub macOS runners to a root EAS workflow. GitHub calculates the stable/canary version, dispatches the exact source SHA to EAS, waits for the package, and publishes the returned tarball using npm OIDC. The GitHub release attaches that same tarball.

The EAS job initializes pinned submodules, installs the root lockfile, applies the requested version, builds all packages, runs tests, and verifies the npm package includes the native helpers and WebRTC framework. Root EAS configuration reuses the existing hub project identity. Version commits now happen after the EAS build succeeds; native caching is deferred.

Validation:

- GitHub CI passed.
- [EAS build from this PR commit](https://expo.dev/accounts/expo/projects/expo-device-hub/workflows/01a08c81-8aff-7135-b728-a544a1045fb7) passed: full macOS build, tests, npm packing, verification, and upload.
- Downloaded that EAS artifact using the release workflow's JSON selection and verified it locally.
- EAS workflow validation, actionlint, and all nine artifact-verification tests passed.
- Local monorepo build/test passed (with existing Turbo cache hits).

No npm publication or release tags were created. The test invoked EAS CLI directly at the pushed commit; the publishing GitHub workflow itself was not dispatched. The existing EAS repository/root linkage works. Before enabling releases, ensure EXPO_DEV_EXPO_GITHUB_ROBOT_ACCESS_TOKEN is available to the GitHub npm-publish job. RELEASING.md documents setup and failure behavior.

Authored by Codex (GPT-6).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,7 +18,8 @@ concurrency:
 
 jobs:
   release:
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
     environment: npm-publish
     env:
       IS_CANARY: ${{ github.event_name == 'push' || inputs.canary }}
@@ -42,15 +43,10 @@ jobs:
         uses: actions/setup-node@v6
         with:
           node-version: "24.x"
+          registry-url: https://registry.npmjs.org
 
       - name: Install dependencies
         run: bun install --frozen-lockfile
-
-      - name: Build packages
-        run: bun run build
-
-      - name: Test
-        run: bun run test
 
       - name: Configure git
         run: |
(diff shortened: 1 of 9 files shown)


.github/workflows/release.yml
```diff
@@ -71,27 +67,58 @@ jobs:
...
-        run: ./scripts/ignore-packages.ts
-
       - name: Publish to npm
         if: ${{ env.IS_CANARY != 'true' }}
         env:
           NPM_CONFIG_PROVENANCE: "true"
-        run: bun run changeset:publish
+          RELEASE_VERSION: ${{ steps.release.outputs.version }}
+        run: |
+          npm publish "release-artifacts/expo-device-hub-${RELEASE_VERSION}.tgz" --access public
```
krystofwoldrich-agent, line 113: **[design]** npm provenance attests a build that did not happen in this workflow.

The publish steps keep `NPM_CONFIG_PROVENANCE: "true"`. npm signs a SLSA statement that says this GitHub workflow run built the tarball from `github.sha`. The tarball is now built on EAS and only downloaded here. npm does not check where the file came from, so the attestation will verify but describe the wrong builder.

This package ships native binaries (`.node`, `.dylib`, the WebRTC framework), which is where consumers rely on provenance most. Either drop `NPM_CONFIG_PROVENANCE` on both publish steps, or state in RELEASING.md that the attestation covers only the publish step and not the build.

scripts/create-github-releases.ts
```diff
@@ -46,9 +48,9 @@ for (const pkg of await getPublicPackages()) {
     if (section) notes = section;
   }
 
-  const packOutput = await $`npm pack --pack-destination ${artifactsDir} --json`.cwd(pkg.dir).text();
-  const tarball = `${artifactsDir}/${JSON.parse(packOutput)[0].filename}`;
+  const tarball = `${artifactsDir}/expo-device-hub-${pkg.version}.tgz`;
```
krystofwoldrich-agent, line 51: **[suggestion]** The loop still iterates every public, non-ignored package, but the tarball path is fixed to `expo-device-hub-<version>.tgz`. If a second package ever becomes public, `verifyReleaseArtifact` throws `Expected public expo-device-hub@...` for it, since the name is hard-coded there too. It fails loudly, so this is not a bug today.

Consider deriving the file name from `pkg.name` (npm's `@scope/name` → `scope-name`), or replace the loop with one explicit package so the code matches what the pipeline supports.

RELEASING.md
```diff
@@ -10,8 +10,41 @@ layer, is marked `private` and is skipped by the release tooling.
...
+commit SHA with that version as an explicit workflow input. EAS initializes the pinned
+submodules and applies the version before building. This also supports canaries without
+pushing temporary version commits. The tarball's published files do not include the changelog;
+GitHub retains the locally generated changelog for its release notes.
+
+The Ubuntu job waits for EAS, downloads the named artifact, and verifies its package name,
+version, web/server output, native helpers, and WebRTC runtime files. Stable releases then
+commit/push the version changes, publish the downloaded tarball, create/push the npm-version
+tag, and attach the same tarball to the GitHub release. Canaries publish the tarball under
+`canary` without creating commits, tags, or GitHub releases. Publishing uses `npm publish`
```
krystofwoldrich-agent, line 34: **[suggestion]** The release no longer calls `changeset publish`. The root `changeset:publish` script and `scripts/ignore-packages.ts` are now unused. The header comment in `scripts/apply-canary-version.ts` ("Run before `changeset publish --tag canary`") and `.changeset/README.md` ("skipped by `changeset publish`") still describe the old flow. Remove the script and update those two comments so they match this section.

krystofwoldrich-agent (changes requested):

**Verdict:** REQUEST_CHANGES
**Findings:** 0 critical · 1 design · 2 suggestion · 0 nit

The PR moves the release build from GitHub macOS runners to an EAS workflow. GitHub computes the version, dispatches EAS at the source SHA, downloads the tarball, verifies it, and publishes with npm OIDC. The pipeline is consistent end to end. The eas-cli 24.0.0 flags used (`--ref`, `--input`, `--wait`, `--json`) exist. The verifier covers the native helpers and the WebRTC framework. The scrcpy server is downloaded at runtime by serve-emu, so it is correctly not a required file.

The main concern is npm provenance. The workflow still sets provenance on, but the build now happens outside the attested workflow.

### Findings

- **[design]** `.github/workflows/release.yml:113` — npm provenance attests a build that did not happen in this workflow. Disable it, or document the limitation.
- **[suggestion]** `scripts/create-github-releases.ts:51` — the loop over public packages hard-codes the hub tarball name.
- **[suggestion]** `RELEASING.md:34` — `changeset:publish`, `scripts/ignore-packages.ts`, and two comments still describe the old publish flow.

Details are in the inline comments.

PR #102, opened 2026-09-10
Add Android video FPS and bitrate controls to Hub UI

krystofwoldrich-agent:
Android stream options exposed only Max size even though serve-emu supports runtime FPS and bitrate changes. Enable the existing Video FPS and Video bitrate controls for Android WebSocket and WebRTC playback, and forward validated FPS/bitrate patches to the server.

Changes use the existing stream restart flow, disable controls while pending, and restore the previous settings if the update fails. Document the controls and add a minor changeset.

Validation:
- 67 targeted tests passed across Android stream contracts, hook restart/rollback behavior, and inspector rendering.
- hub-client and hub-components builds passed.
- hub-client typecheck and lint passed; git diff --check passed.
- Live device playback was not tested.

PR #98, opened 2026-09-10
Default Android streaming to RGB888

krystofwoldrich-agent:
Android streams now use RGB888 over gRPC by default in both the standalone CLI and the Expo CLI plugin, replacing MMAP. Explicit PNG, MMAP, and scrcpy selections remain available.

Updates the existing default-option tests and CLI help expectations, adds a minor changeset, and removes the Android streaming section from the README.

Validation: all 177 Device Hub tests pass; the server and CLI bundles build successfully. The follow-up documentation and changeset edits pass git diff --check.

PR #96, opened 2026-09-10
Add Android hardware encoding controls to Device Hub

krystofwoldrich:
Adds Android-only `--encoder software|hardware`, Expo plugin JSON options, client status/setter plumbing and an Encoder row in Stream options. Software remains the default. The UI reports the active FFmpeg backend and shows one error alert when a switch fails, while the current capture keeps streaming. Hardware stays selectable for retry. Plugin JSON encoder values are normalized and validated at startup.

Includes a minor changeset and a reusable [browser comparison recipe and animation fixture](https://github.com/expo/expo-device-hub/blob/codex/android-encoder-hub/packages/serve-emu/packages/serve-emu/docs/hardware-encoder-spike.md#reproducing-the-browser-comparison).

Validation: all 399 Hub package tests pass, including regression coverage for duplicate alerts, server-only advisory errors, structured API errors, unrelated client failures and invalid plugin JSON encoder values. The combined stack passes 973 serve-emu tests, critical coverage, typechecks, lint, builds, docs checks and package smoke. All three latest Ubuntu CI checks pass.

The final built CLI was tested in the Codex browser with WebRTC and a 60 FPS cap. A temporary FFmpeg wrapper failed exactly one hardware smoke encode: the UI showed one detailed alert, Hardware remained selectable, and Software kept streaming at generation 0. Retrying without restarting the server resumed live VideoToolbox video at generation 1 and cleared the alert. The normal server was then restarted with hardware, WebRTC and the same 60 FPS cap.

Initial browser measurements on 2026-09-10 used Apple M4 Pro, macOS 26.6.2, FFmpeg 9.0.1 and the Codex in-app browser. These are observations from that host and software version, not portable performance guarantees:

- Live WebSocket and WebRTC H.264 video used `h264_videotoolbox`; API diagnostics and the running server process confirmed `-allow_sw 0` and `-flags +low_delay`.
- Software ↔ Hardware switching, refresh, late joining and continued decoding after rotation worked. Hardware switching took approximately 1.98 seconds, refresh 1.17 seconds and late joining 2.21 seconds.
- Matched 15.6-second motion windows at 540×1170 / MMAP / 30 FPS cap / 8 Mbps measured FFmpeg CPU at 12.14% software and 5.51% hardware, comparable source throughput around 23 FPS, and zero server frame drops. CPU is accumulated process CPU time divided by wall time; one occupied core is 100%.
[...]

PR #108, opened 2026-09-11
Select devices by URL and retain offline device views

krystofwoldrich-agent:
Device selection now follows `/device/{udid}` for iOS and `/device/{serial}` for Android, including direct links and browser history. If the selected device disappears from discovery, its row, frame, and options remain visible with an offline message until the user selects another device.

Discovery snapshots reuse unchanged device objects, and list/viewer subscriptions are separated so unrelated device updates do not rerender the viewer or unchanged rows.

Validation: full Device Hub build; regression tests for URL navigation, snapshot reconciliation, render isolation, and offline frame behavior. The combined branch with the dependent boot/create work passes 296 dashboard, server, and component tests, plus component typecheck and lint.

Follow-up: [#109 — Show device creation and boot progress outside the dialog](https://github.com/expo/expo-device-hub/pull/109). Merge this PR first, then rebase #109 onto main. Both target main because the branches live in a fork and cannot use GitHub stacks here.

### DIFF
--- a/packages/@expo/hub-components/src/__tests__/dashboard-device-updates.test.tsx
+++ b/packages/@expo/hub-components/src/__tests__/dashboard-device-updates.test.tsx
@@ -1,0 +1,215 @@
+import '../../../../expo-device-hub/types.d.ts';
+
+import { expect, spyOn, test } from 'bun:test';
+import * as hubClient from '@expo/hub-client';
+import { act, type ReactPortal } from 'react';
+import * as ReactDOM from 'react-dom';
+import { create, type ReactTestRenderer } from 'react-test-renderer';
+
+import Dashboard from '../../../../expo-device-hub/src/Dashboard';
+import { DEVICE_LIST_MESSAGE_TYPE } from '../../../../expo-device-hub/src/device-list-protocol';
+import { useDashboardStore } from '../../../../expo-device-hub/src/dashboard/dashboardStore';
+import { useDeviceSessionStore } from '../../../../expo-device-hub/src/dashboard/deviceSessionStore';
+import { type DeviceList } from '../../../../expo-device-hub/src/dashboard/useDevices';
+import { NOOP_DEVICE_CLIENT } from '../../../hub-client/src/useNoopDeviceClient';
+import { type Device } from '../dashboard/data';
+
+const IPHONE: Device = {
+  id: 'A671E25C-3A15-4738-8B49-8FACD3AE5ACD',
+  name: 'Opened iPhone',
+  version: 'iOS 27.0',
+  platform: 'ios',
+  booted: true,
+  physical: false,
+  supported: true,
+  deviceFrame: 'ios:iphone-17-pro',
+};
+
+class FakeWebSocket {
+  static sockets: FakeWebSocket[] = [];
+  onopen: (() => void) | null = null;
+  onmessage: ((event: { data: string }) => void) | null = null;
+  onerror: (() => void) | null = null;
+  onclose: (() => void) | null = null;
+
+  constructor(readonly url: string) {
+    FakeWebSocket.sockets.push(this);
+  }
+
+  close() {}
+
+  snapshot(devices: DeviceList) {
+    this.onmessage?.({ data: JSON.stringify({ type: DEVICE_LIST_MESSAGE_TYPE, devices }) });
+  }
+}
+
+function browserEnvironment(path: string) {
+  const events = new EventTarget();
+  let location = new URL(path, 'http://localhost:8081');
+  const browser = {
+    __EXPO_DEVICE_HUB_BASE_PATH__: '',
+    innerWidth: 1440,
+    isSecureContext: true,
+    localStorage: { getItem: () => null, setItem: () => {} },
+    get location() {
+      return location;
+    },
+    history: {
+      state: null,
+      pushState(_state: unknown, _unused: string, url: string) {
+        location = new URL(url, location);
+      },
+      replaceState(_state: unknown, _unused: string, url: string) {
+        location = new URL(url, location);
+      },
+    },
+    matchMedia: (query: string) => ({
+      matches: query.includes('prefers-reduced-motion'),
+      addEventListener() {},
+      removeEventListener() {},
+    }),
+    setTimeout,
+    clearTimeout,
+    addEventListener: events.addEventListener.bind(events),
+    removeEventListener: events.removeEventListener.bind(events),
+    dispatchEvent: events.dispatchEvent.bind(events),
+  };
+  const replacements = {
+    window: browser,
+    document: {
+      body: {},
+      documentElement: { classList: { toggle() {}, remove() {} } },
+    },
+    WebSocket: FakeWebSocket,
+    IS_REACT_ACT_ENVIRONMENT: true,
+  };
+  const descriptors = new Map(
+    Object.keys(replacements).map((key) => [key, Object.getOwnPropertyDescriptor(globalThis, key)])
+  );
+  for (const [key, value] of Object.entries(replacements)) {
+    Object.defineProperty(globalThis, key, { configurable: true, writable: true, value });
+  }
+  return {
+    browser,
+    restore() {
+      for (const [key, descriptor] of descriptors) {
+        if (descriptor) Object.defineProperty(globalThis, key, descriptor);
+        else Reflect.deleteProperty(globalThis, key);
+      }
+    },
+  };
+}
+
+test('dashboard isolates device-list updates and retains the selected offline frame', async () => {
+  const environment = browserEnvironment(`/device/${IPHONE.id}`);
+  const previousSession = useDeviceSessionStore.getState();
+  const previousDashboard = useDashboardStore.getState();
+  useDeviceSessionStore.setState(useDeviceSessionStore.getInitialState(), true);
+  useDashboardStore.setState(useDashboardStore.getInitialState(), true);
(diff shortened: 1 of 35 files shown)


packages/expo-device-hub/src/Dashboard.tsx
```diff
@@ -260,40 +164,39 @@ export default function Dashboard(_props: { dom?: import('expo/dom').DOMProps })
...
+    [selected?.platform, streamModeAvailability]
+  );
+  const handleStreamModeChange = useCallback(
+    (mode: DeviceStreamMode) => {
+      chooseStreamMode(mode, selectedStreamModeAvailability);
+    },
+    [chooseStreamMode, selectedStreamModeAvailability]
+  );
+  const handleShutdown = useCallback(async () => {
+    if (selected) await shutdownDevice(selected);
```
krystofwoldrich-agent, line 184: **[design]** Shut down / Remove no longer move the selection.

Before this PR `handleShutdown` / `handleRemove` called `dismissDevice(id)`, which cleared the selection so `reconcileSelectedDevice` moved to the next running device. Now the handlers only post the action. The URL still names the device, so `reconcileDeviceSession` retains it: the row shows "Offline", the frame shows `UnavailableDeviceScreen`, and every inspector action (including Shut down and Remove) is disabled. For Remove this is incorrect: the device no longer exists, yet `PhoneFrame.tsx:255` tells the user "Its screen will return when it reconnects."

Offline retention fits an unexpected disappearance; user-initiated shutdown/remove is a known transition. Suggest: after a successful `shutdownDevice` / `removeDevice`, call `selectDeviceRoute(nextRunningId ?? '', { replace: true })`, or clear the route and let `DeviceDiscovery` pick the first device. `dashboard-device-updates.test.tsx:177` covers only the unexpected-removal path, so this regression is untested.

packages/@expo/hub-components/src/__tests__/dashboard-device-updates.test.tsx
```diff
@@ -0,0 +1,215 @@
+import '../../../../expo-device-hub/types.d.ts';
+
+import { expect, spyOn, test } from 'bun:test';
+import * as hubClient from '@expo/hub-client';
+import { act, type ReactPortal } from 'react';
+import * as ReactDOM from 'react-dom';
+import { create, type ReactTestRenderer } from 'react-test-renderer';
+
+import Dashboard from '../../../../expo-device-hub/src/Dashboard';
```
krystofwoldrich-agent, line 9: **[design]** This test renders `expo-device-hub`'s `Dashboard`, `dashboardStore`, `deviceSessionStore`, and `device-list-protocol` via `../../../../expo-device-hub/src/...`; `device-section-renders.test.tsx:8` does the same for `reconcileDeviceList`. `@expo/hub-components` is documented (its `package.json` description and `packages/expo-device-hub/CLAUDE.md`) as the dependency-free UI kit shared with the website; it should not depend on the app. Line 1 (`import '../../../../expo-device-hub/types.d.ts'`) exists only so this package's `tsc` accepts the app's `.css` / `.png` ambient modules, which confirms the inversion.

The likely cause is that `react-test-renderer` was added as a devDependency here (`package.json:62`) rather than in `expo-device-hub`, whose existing component tests (`inspectorSections.test.tsx`, `bootDeviceVisibility.test.tsx`) use `react-dom/server`. Move both tests to `packages/expo-device-hub/src/dashboard/__tests__/` and add the renderer devDependency there. Separately, `react-test-renderer` is deprecated as of React 19 and prints a deprecation warning on use; `@testing-library/react` or `react-dom/client` + `act` avoids taking on a dead dependency.

packages/@expo/hub-components/src/dashboard/LogSidebar.tsx
```diff
@@ -62,6 +63,20 @@ export function LogSidebar({
...
+  useEffect(() => {
+    if (available && device && client) {
+      lastClient.current = { deviceId: device.id, client };
+    } else if (lastClient.current?.deviceId !== device?.id) {
+      lastClient.current = null;
+    }
+  }, [available, client, device]);
+  // The connection hook resets on disconnect. Retain the selected device's
+  // metadata and values so the same inspector sections and rows stay in place.
+  const inspectorClient =
```
krystofwoldrich-agent, line 76: **[design]** "Last live value" retention is duplicated in two components via refs read during render.

`LogSidebar` caches the last connected `client` in `lastClient` (lines 66-79) and `PhoneFrame` caches the last `screen` in `lastScreen` (`PhoneFrame.tsx:89-106`). Both write the ref in an effect and read it during render, so the rendered output depends on the previous commit's effect rather than on props/state; under StrictMode double-render or a concurrent re-render this can show one frame of the reset (`NOOP_DEVICE_CLIENT`) before the ref catches up.

The session store already owns the "retained selected device" concept (`selectedDevice` when `selectedAvailable` is false). A single retention point, e.g. `DashboardLayout` keeping the last non-noop `client` for `selected.id` in state and passing it as the inspector client, or the store holding `lastScreen`, would remove both caches and make the offline view a pure function of props.

packages/expo-device-hub/src/dashboard/deviceSessionStore.ts
```diff
@@ -0,0 +1,112 @@
...
+    emulators: EMPTY_DEVICES,
+    recent: { simulators: EMPTY_DEVICES, emulators: EMPTY_DEVICES },
+    selectedAvailable: false,
+    connectionStatus: 'connecting',
+    update: (update) => set((state) => reconcileDeviceSession(state, update)),
+    rememberDevice: (selectedDevice) =>
+      set((state) =>
+        state.selectedAvailable && state.selectedDevice?.id === selectedDevice.id
+          ? state
+          : { selectedDevice, selectedAvailable: false }
```
krystofwoldrich-agent, line 106: **[suggestion]** `rememberDevice` flips `selectedAvailable` to false as a side effect.

`handleAddDevice` (`Dashboard.tsx:476`) calls `rememberDevice(device)` before `selectDeviceRoute(device.id)`. For any device other than the currently selected online one, this synchronously sets `selectedDevice` to the new device and `selectedAvailable: false` while `simulators` / `emulators` still describe the old lists. The result is one render in which `useActiveDeviceClient` receives `null`, `StreamPanel` shows the new device as offline, and `DeviceSidebar` computes `offlineDeviceId` for a device that has no row yet.

`reconcileDeviceSession` already falls back to `previous.selectedDevice` when the id matches; the only value `rememberDevice` adds is metadata for a brand-new device before discovery. A `knownDevices: Record<string, Device>` map consulted at line 58 (before `unknownDevice`) would provide the same metadata without touching selection or availability.

krystofwoldrich-agent (commented):

## Review: Select devices by URL and retain offline device views

**Verdict:** REQUEST_CHANGES (posted as a comment because the reviewer account owns this PR)
**Findings:** 0 critical · 3 design · 1 suggestion · 3 nit

PR 1/2 in the stack (#109 builds on this branch). The URL routing in `deviceRoute.ts`, the server shell route, and the snapshot reconciliation are sound. I checked the encoding path end to end (`encodeURIComponent` → `location.pathname` → `readDeviceRoute`, server regex on the still-encoded pathname) and the identity-preservation claims in `reconcileDeviceList` / `reconcileDeviceSession`; both hold for the `Device` type, which has only primitive fields. No old consumers of `selectedDeviceId` / `addedDevices` / `dismissDevice` remain. The main problems are behavioral: explicit Shut down / Remove now leave the user on a dead device with a misleading message, and new Dashboard-level tests live in the shared UI kit package and reach into the app package by relative path. The PR also adds no `.changeset/*.md` entry; `RELEASING.md` requires one for anything that should ship, and the recent feature PRs (#104, #96, #102, #101, #98, #80) each include one.

### Findings

- **[design]** Shut down / Remove no longer move the selection; the removed device stays selected as "offline" with a message saying it will return. `packages/expo-device-hub/src/Dashboard.tsx:184`
- **[design]** Dashboard integration tests live in `@expo/hub-components` and import the app package by relative path. `packages/@expo/hub-components/src/__tests__/dashboard-device-updates.test.tsx:9`
- **[design]** "Last live value" retention is duplicated in `LogSidebar` and `PhoneFrame` via refs read during render. `packages/@expo/hub-components/src/dashboard/LogSidebar.tsx:76`
- **[suggestion]** `rememberDevice` flips `selectedAvailable` to false as a side effect. `packages/expo-device-hub/src/dashboard/deviceSessionStore.ts:106`
- **[nit]** Redundant `usedByAgent` guard: `DeviceListItem` already masks the agent status with `active={!offline && usedByAgent}` and builds the aria-label from `offline` first, so the `device.id !== offlineDeviceId &&` prefix duplicates that rule. `packages/@expo/hub-components/src/dashboard/DeviceSection.tsx:119`
- **[nit]** Unused deselect branch: `selectDeviceRoute('')` has no production caller, only `deviceRoute.test.ts` exercises it, and `DeviceDiscovery` would immediately re-select the first device. Either drop it or use it for the shutdown/remove flow above. `packages/expo-device-hub/src/dashboard/deviceRoute.ts:44`
- **[nit]** `Object.assign(globalThis, { IS_REACT_ACT_ENVIRONMENT: true })` at module scope leaks into every other file in the same `bun test` process. The sibling tests save and restore the previous value; this file should too. `packages/@expo/hub-components/src/__tests__/offline-device.test.tsx:70`

Details are in the inline comments. Tests were not run for this review: the checkout had no full `node_modules`. Conclusions come from reading the tests against the implementation.

PR #109, opened 2026-09-11
Show device creation and boot progress outside the dialog

krystofwoldrich-agent:
Boot and Create now immediately add and select a local device. Users can close the dialog while the operation continues; the sidebar and phone frame show Creating… or Booting…, then connect when discovery confirms the device is online. Failures remain in the frame with the error and a Try again button.

Creation and boot are separate steps so a failed boot retries the existing device. The create API accepts optional `boot: false` while preserving its default create-and-boot behavior. Local-to-real device IDs are reconciled without duplicate Android rows or navigating away from another selected device. A dismissed dialog's response cannot close a newly opened dialog.

Depends on #108 and includes its commits. Merge #108 first, then rebase this PR onto main. Both target main because the branches live in a fork and cannot use GitHub stacks here.

[Base PR: URL selection and offline views](https://github.com/expo/expo-device-hub/pull/108) · [Focused diff for this follow-up](https://github.com/krystofwoldrich-agent/expo-device-hub/compare/codex/device-url-selection...codex/device-start-progress)

Validation: full Device Hub build; 296 passing dashboard, server, and component tests; component typecheck and lint; `git diff --check`. Tests cover immediate state updates, dismiss/reopen races, creation and boot failures, retries, Android ID changes, discovery reconciliation, and render isolation.

### DIFF
--- a/packages/@expo/hub-components/src/__tests__/device-start-dialog.test.tsx
+++ b/packages/@expo/hub-components/src/__tests__/device-start-dialog.test.tsx
@@ -1,0 +1,223 @@
+import { expect, mock, spyOn, test } from 'bun:test';
+import { act } from 'react';
+import { create, type ReactTestRenderer } from 'react-test-renderer';
+import { renderToStaticMarkup } from 'react-dom/server';
+
+import * as dialog from '../components/Dialog';
+import { type AddDeviceOutcome, type Device, type NewDeviceOptions } from '../dashboard/data';
+import { RecentDevicesModal } from '../dashboard/RecentDevicesModal';
+import { PhoneFrame } from '../dashboard/PhoneFrame';
+import { DeviceTitle } from '../dashboard/DeviceTitle';
+import { DeviceSection } from '../dashboard/DeviceSection';
+import { createDeviceStarter } from '../../../../expo-device-hub/src/dashboard/startDevice';
+import { createDeviceSessionStore } from '../../../../expo-device-hub/src/dashboard/deviceSessionStore';
+import { type StartDeviceOutcome } from '../../../../expo-device-hub/src/dashboard/deviceActions';
+
+const device: Device = {
+  id: 'ios-udid',
+  name: 'iPhone',
+  version: 'iOS 27',
+  platform: 'ios',
+  booted: false,
+  physical: false,
+  supported: true,
+  deviceFrame: 'ios:iphone-17-pro',
+};
+const options: NewDeviceOptions = { runtimes: [] };
+function deferred<T>() {
+  let resolve!: (value: T) => void;
+  const promise = new Promise<T>((done) => {
+    resolve = done;
+  });
+  return { promise, resolve };
+}
+
+test('closing the boot dialog leaves an immediate local device and its later failure', async () => {
+  const environment = globalThis as typeof globalThis & { IS_REACT_ACT_ENVIRONMENT?: boolean };
+  const previousEnvironment = environment.IS_REACT_ACT_ENVIRONMENT;
+  environment.IS_REACT_ACT_ENVIRONMENT = true;
+  // Only replace the DOM portal; exercise the real form and its submission lifecycle.
+  const content = spyOn(dialog, 'DialogContent').mockImplementation(({ children }) => (
+    <>{children}</>
+  ));
+  const store = createDeviceSessionStore();
+  const booted = deferred<StartDeviceOutcome>();
+  let route = '';
+  const start = createDeviceStarter({
+    store,
+    boot: () => booted.promise,
+    create: async () => {
+      throw new Error('Unexpected create');
+    },
+    navigate: (id) => {
+      route = id;
+    },
+    currentRoute: () => route,
+    newRequestId: () => 'request',
+  });
+  const close = mock(() => {});
+  let renderer: ReactTestRenderer | undefined;
+  let submitted!: Promise<void>;
+  try {
+    await act(() => {
+      renderer = create(
+        <RecentDevicesModal
+          open
+          kind="simulator"
+          devices={[device]}
+          options={options}
+          onAdd={start}
+          onClose={close}
+        />
+      );
+    });
+    await act(() => {
+      submitted = renderer!.root.findByType('form').props.onSubmit({ preventDefault() {} });
+    });
+    expect(store.getState().simulators).toHaveLength(1);
+    expect(store.getState().selectedDevice?.startup?.phase).toBe('booting');
+    const cancel = renderer!.root
+      .findAllByType('button')
+      .find((button) =>
+        button.children.some(
+          (child) =>
+            typeof child === 'object' && 'children' in child && child.children.includes('Close')
+        )
+      );
+    expect(cancel).toBeDefined();
+    expect(cancel!.props.disabled).toBeFalsy();
+    await act(() => {
+      renderer!.root.findByType(dialog.DialogRoot).props.onOpenChange(false);
+    });
+    expect(close).toHaveBeenCalledTimes(1);
+    await act(async () => {
+      booted.resolve({ id: null, error: 'Unable to start: device is unavailable' });
+      await submitted;
+    });
+    expect(store.getState().selectedDevice?.startup).toEqual({
+      phase: 'failed',
+      action: 'boot',
+      message: 'Unable to start: device is unavailable',
+    });
+    expect(close).toHaveBeenCalledTimes(1);
+  } finally {
+    await act(() => renderer?.unmount());
+    content.mockRestore();
+    environment.IS_REACT_ACT_ENVIRONMENT = previousEnvironment;
+  }
+});
+
+test.each([true, false])(
+  'a dismissed request (success=%s) cannot affect a reopened dialog',
+  async (success) => {
+    const environment = globalThis as typeof globalThis & { IS_REACT_ACT_ENVIRONMENT?: boolean };
+    const previousEnvironment = environment.IS_REACT_ACT_ENVIRONMENT;
+    environment.IS_REACT_ACT_ENVIRONMENT = true;
+    const content = spyOn(dialog, 'DialogContent').mockImplementation(({ children }) => (
+      <>{children}</>
+    ));
+    const first = deferred<AddDeviceOutcome>();
+    const second = deferred<AddDeviceOutcome>();
+    const onAdd = mock().mockReturnValueOnce(first.promise).mockReturnValueOnce(second.promise);
+    const close = mock(() => {});
+    const view = (open: boolean) => (
+      <RecentDevicesModal
+        open={open}
+        kind="simulator"
+        devices={[device]}
+        options={options}
+        onAdd={onAdd}
+        onClose={close}
+      />
+    );
+    let renderer: ReactTestRenderer | undefined;
+    let firstSubmission!: Promise<void>;
+    let secondSubmission!: Promise<void>;
+    try {
+      await act(() => {
+        renderer = create(view(true));
+      });
+      await act(() => {
+        firstSubmission = renderer!.root.findByType('form').props.onSubmit({ preventDefault() {} });
+      });
+      await act(() => renderer!.update(view(false)));
+      await act(() => renderer!.update(view(true)));
+      await act(() => {
+        secondSubmission = renderer!.root
+          .findByType('form')
+          .props.onSubmit({ preventDefault() {} });
+      });
+      await act(async () => {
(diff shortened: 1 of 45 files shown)


packages/expo-device-hub/src/dashboard/deviceSessionStore.ts
```diff
@@ -0,0 +1,202 @@
...
+): DeviceSession {
+  const { booted, recent, selectedId, platform } = update;
+  const running = [...booted.simulators, ...booted.emulators];
+  const matches = (local: Device, device: Device) =>
+    local.platform === device.platform &&
+    (local.id === device.id || (local.platform === 'android' && local.name === device.name));
+  const retained = Object.values(previous.startups).filter(
+    (entry) =>
+      entry.pending ||
+      entry.device.startup?.phase === 'failed' ||
```
krystofwoldrich-agent, line 70: **[critical]** A failed startup entry hides the same device once it comes online.

The `retained` filter keeps every entry whose `startup.phase === 'failed'`, regardless of `running`. `displayed` (lines 80-83) and `withLocalDevices` (line 99) then remove any running device that `matches` the local entry by id or by Android AVD name, and `selectedAvailable` (line 94) is `false` because the selected device carries `startup`. Result: the sidebar shows "Boot failed", the frame shows the error, and `useActiveDeviceClient` in `Dashboard.tsx:190-195` never connects, even though discovery lists the device as booted.

This is reachable in normal use. `bootAndroidHubDevice` returns `ok: false` after `BOOT_READY_TIMEOUT_MS` (`server/device-actions.ts:374`), and a slow emulator can come online after that. For a recent Android device the local id is the AVD name (`server/devices.ts:110`), so the name-based `matches` clause hides the emulator's serial row. The same applies to iOS when `simctl boot` fails for a device that is already booted. The only exit is "Try again", which then short-circuits through `alreadyRunning` (`startDevice.ts:110-119`).

Reproduced in a scratch copy: start a recent device, resolve `boot` with an error, then push a snapshot with `{ ...iphone, booted: true }`. `selectedAvailable` stays `false` and `simulators[0].startup.phase` stays `'failed'`.

Fix: drop a failed entry when `running` contains a match (id, or Android name), the same rule already applied to completed entries on lines 71-73. Add the scenario as a test in `startDevice.test.ts`.

packages/@expo/hub-components/src/dashboard/data.ts
```diff
@@ -42,6 +53,8 @@ export type Device = {
    * add-device picker.
    */
   lastUsedAt?: number;
+  /** Present only on locally requested devices while starting or after a failure. */
+  startup?: DeviceStartup;
```
krystofwoldrich-agent, line 57: **[design]** Browser-only lifecycle state is added to the shared `Device` model.

`Device` is the data shape shared with the website (`AGENTS.md`: components were extracted "so the website can consume the same code"). #108 modelled viewer-local state as separate props (`offlineDeviceId` on `DeviceSection`, `available` on `PhoneFrame` / `StreamPanel`, `offline` on `DeviceListItem`). This PR instead embeds `startup` in `Device`, so `DeviceTitle`, `PhoneFrame`, `DeviceSection` and `DeviceListItem` now branch on a field that only the Hub dashboard ever populates.

The field also leaks into the retry target: `createdDevice` at `startDevice.ts:99-107` includes `startup: { phase: 'booting' }` and becomes `target.device`, while the success path at line 131 strips it. So `AddDeviceTarget.device` sometimes carries a `startup` value that this comment says is "present only on locally requested devices".

Either keep the field and document it as dashboard-owned (and strip it consistently at `startDevice.ts:107`), or pass startup state as a prop keyed by id, following the `offlineDeviceId` precedent.

packages/@expo/hub-components/src/dashboard/PhoneFrame.tsx
```diff
@@ -186,3 +217,88 @@ export function PhoneFrame({
...
+            color: text.secondary,
+            margin: 0,
+            whiteSpace: 'pre-wrap',
+            overflowWrap: 'anywhere',
+            maxHeight: '35cqh',
+            overflowY: 'auto',
+          }}>
+          {message}
+        </p>
+        {failed && onRetry && (
```
krystofwoldrich-agent, line 296: **[design]** Failed rows cannot be dismissed.

A failed entry is retained until the same target is started again (`startDevice.ts:59-72`) or the page reloads. "Try again" is the only control. For a failed create with a placeholder id (`pending-…`) there is no matching recent, so the row is a permanent ghost in the sidebar. Example: creation fails because the AVD name already exists; retry repeats the failure; the user cannot remove the row.

Add a `removeStartup(requestId)` store action and a "Dismiss" control next to "Try again". This also gives the user an exit for the critical finding in `deviceSessionStore.ts:70` until it is fixed.

packages/expo-device-hub/src/dashboard/startDevice.ts
```diff
@@ -0,0 +1,150 @@
...
+          (device.platform === 'android' && item.device.name === device.name) ||
+          (target.kind === 'new' &&
+            item.target.kind === 'new' &&
+            item.target.device.name === target.device.name &&
+            item.target.device.runtime === target.device.runtime &&
+            item.target.device.deviceType === target.device.deviceType))
+    );
+    if (failed) entry.requestId = failed.requestId;
+    store.getState().putStartup(entry, true);
+    navigate(device.id);
```
krystofwoldrich-agent, line 74: **[design]** Placeholder ids are written to the URL and browser history.

`navigate(device.id)` pushes `/device/pending-<hex>` (id built on line 48). Within the session the alias map redirects it (`DeviceDiscovery.tsx:15-19`). After a reload, or when the URL is shared, the id is unknown. `unknownDevice` (`deviceSessionStore.ts:43-55`) then guesses `android` for any non-UUID id and renders a "Device / pending-…" offline row, also for an iOS creation. History entries from `publish` (line 82) and from retries (this line again, with a new id) accumulate.

Options: select the placeholder in the store only and navigate when a real id exists; or treat an unresolvable `pending-` route as no selection in `DeviceDiscovery`.

packages/expo-device-hub/src/dashboard/deviceSessionStore.ts
```diff
@@ -0,0 +1,202 @@
...
+  update: SessionUpdate
+): DeviceSession {
+  const { booted, recent, selectedId, platform } = update;
+  const running = [...booted.simulators, ...booted.emulators];
+  const matches = (local: Device, device: Device) =>
+    local.platform === device.platform &&
+    (local.id === device.id || (local.platform === 'android' && local.name === device.name));
+  const retained = Object.values(previous.startups).filter(
+    (entry) =>
+      entry.pending ||
```
krystofwoldrich-agent, line 69: **[suggestion]** No upper bound for "Booting…" after HTTP success.

After `boot` resolves, the entry has `pending: false` and `startup: booting` (`startDevice.ts:121-132`). It is retained until `running` contains the id (lines 71-73). If discovery never lists the device, the row shows "Booting…" indefinitely and "Try again" is not offered. iOS discovery filters simulators by `lastUsedAt !== undefined` (`server/devices.ts:95`), so listing depends on usage history being present. Consider converting the entry to `failed` after a timeout with a clear message, so the user has a retry path.

packages/@expo/hub-components/src/__tests__/device-start-dialog.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+import { create, type ReactTestRenderer } from 'react-test-renderer';
+import { renderToStaticMarkup } from 'react-dom/server';
+
+import * as dialog from '../components/Dialog';
+import { type AddDeviceOutcome, type Device, type NewDeviceOptions } from '../dashboard/data';
+import { RecentDevicesModal } from '../dashboard/RecentDevicesModal';
+import { PhoneFrame } from '../dashboard/PhoneFrame';
+import { DeviceTitle } from '../dashboard/DeviceTitle';
+import { DeviceSection } from '../dashboard/DeviceSection';
+import { createDeviceStarter } from '../../../../expo-device-hub/src/dashboard/startDevice';
```
krystofwoldrich-agent, line 12: **[suggestion]** Lines 12-14 import `../../../../expo-device-hub/src/dashboard/{startDevice,deviceSessionStore,deviceActions}`. This inverts the package dependency: hub-components is the shared library and should not depend on the dashboard. `tsconfig.json` includes `src`, so `tsc --noEmit` in hub-components now compiles expo-device-hub source with hub-components' compiler options, and `bun test` in this package alone requires the sibling package. Move the first test to `expo-device-hub/src/dashboard/__tests__/` (which already depends on hub-components), and keep the modal tests here with a fake `onAdd`. Same concern as raised on #108.

packages/expo-device-hub/src/server/device-actions.ts
```diff
@@ -90,12 +92,13 @@ export async function parseCreateDeviceAction(
...
 
   if (!data || typeof data !== 'object') return null;
-  const { platform, name, runtime, deviceType } = data as Record<string, unknown>;
+  const { platform, name, runtime, deviceType, boot } = data as Record<string, unknown>;
   if (
     (platform !== 'ios' && platform !== 'android') ||
     !isNonEmptyString(name) ||
     !isNonEmptyString(runtime) ||
     !isNonEmptyString(deviceType) ||
+    (boot !== undefined && typeof boot !== 'boolean') ||
```
krystofwoldrich-agent, line 101: **[suggestion]** A body with `boot: "false"` is rejected here, but the route answers `Expected { platform, name, runtime, deviceType } JSON body` (`server/index.ts:168`). The message does not say which field failed or that `boot` must be a boolean. Update the message, or return a field-specific error.

krystofwoldrich-agent (commented):

## Review: Show device creation and boot progress outside the dialog

**Verdict:** REQUEST_CHANGES (posted as a comment because the reviewer account owns this PR)
**Findings:** 1 critical · 3 design · 3 suggestion · 2 nit

PR 2/2 in the stack; this review covers only the commit on top of #108 (`b8f12e8a..0e5a4b6a`). The background-start flow is well structured: `createDeviceStarter` owns the operation, the store reconciles local rows against discovery, and the dialog no longer owns the result. The tests cover the main races (dismiss/reopen, Android serial reconciliation, retry of create vs. boot). One reconciliation rule is wrong: a failed local entry is kept forever and shadows the real device when discovery later reports it running, which blocks streaming until the user clicks "Try again". The remaining findings concern the shared data model, sticky failure rows, and placeholder ids in the URL. Like #108, this PR adds no `.changeset/*.md` entry.

### Findings

- **[critical]** A failed startup entry hides the same device once it comes online. `packages/expo-device-hub/src/dashboard/deviceSessionStore.ts:70`
- **[design]** Browser-only lifecycle state (`startup`) is added to the shared `Device` model. `packages/@expo/hub-components/src/dashboard/data.ts:57`
- **[design]** Failed rows cannot be dismissed. `packages/@expo/hub-components/src/dashboard/PhoneFrame.tsx:296`
- **[design]** Placeholder `pending-…` ids are written to the URL and browser history. `packages/expo-device-hub/src/dashboard/startDevice.ts:74`
- **[suggestion]** No upper bound for "Booting…" after HTTP success. `packages/expo-device-hub/src/dashboard/deviceSessionStore.ts:69`
- **[suggestion]** Integration test in hub-components imports expo-device-hub source. `packages/@expo/hub-components/src/__tests__/device-start-dialog.test.tsx:12`
- **[suggestion]** The 400 response does not mention the new `boot` field. `packages/expo-device-hub/src/server/device-actions.ts:101`
- **[nit]** `entry.requestId = failed.requestId` mutates the entry after `device.id` was built from the new `requestId` on line 48, so the store key and the `pending-…` suffix no longer match. It works, but the alias chain is harder to read. Compute the key before building the device, or leave a comment. `packages/expo-device-hub/src/dashboard/startDevice.ts:72`
- **[nit]** `maxHeight: '35cqh'` resolves against the nearest `container-type: size` ancestor. The framed `screenStyle` sets that (`deviceFrame.ts:196`), but the unframed branch (`PhoneFrame.tsx:166-172`) does not, and the wrapper on line 122 is `inline-size` only. In that path `cqh` falls back to the viewport height. `packages/@expo/hub-components/src/dashboard/PhoneFrame.tsx:291`

Details are in the inline comments.

Tests run: `bun test` on the five PR test files in a scratch copy of the branch (dependencies installed, `hub-android-utils`, `hub-apple-utils` and `hub-client` built): 30 pass, 0 fail. A one-off repro test for the critical finding failed as described (`selectedAvailable` stays false after a running snapshot). `tsc --noEmit` did not complete in the copy because other workspace packages lack `dist` output; not attributable to the PR.

PR #112, opened 2026-09-16
Add single-device Android session recording

vonovak:
## Why

Android EAS sessions need a screen recording the session page can play next to timestamped automation events, without an Android-specific player. The recording must survive a killed or stalled Hub and must never stop a session from starting.

## How

- Record H.264 from the canonical serve-emu capture, before browser fanout, with or without a viewer.
- Mux a fragmented MP4 with frame timing, idle gaps and a wall-clock first-frame timestamp. Fragments close at keyframes and reach the disk unbuffered, so a killed Hub leaves a `recording.mp4.partial` that plays up to the last keyframe.
- Write `session.json` at the first keyframe with the file name, first-frame time and size. Finish and fail overwrite it, and failed manifests still name the partial file, so EAS can upload what reached the disk with the reason.
- A capture failure no longer ends the recording. The Hub logs it with the scrcpy stderr tail, restarts the capture on the next request, and the recording continues when the encoder configuration is unchanged; the gap is a held frame. A changed encoder configuration or display size still fails it.
- Finish and shutdown are idempotent. With zero or several booted emulators the Hub starts without recording and logs a warning.
- Limits: 4 GiB file size, 256 MiB free-space reserve, one-hour duration that finalizes cleanly, 30 s finalization bound. Env overrides are validated before recording starts.
- EAS contract: CLI opt-in, `POST /_eas/android-recording/stop` beside `/readyz` with a bearer token, 409 with the reason when nothing was recorded, the existing `recordings.json` / `session.json` upload format. Documented in the README.

Companion EAS stack: https://github.com/expo/eas-cli/pull/4418. Publish the Hub release before deploying the worker.

Scope: one booted emulator, no audio. Rotation and codec changes fail the recording; the prefix is uploaded as a partial.

## Follow-ups

- iOS parity: `record-sim` writes its index at finalize and its manifest at stop, so an iOS crash loses the recording.
- Keyframe cadence bounds the crash loss; on an idle gRPC stream it stretches to minutes.
- Website: show `partial` and `partialReason`.
- A second SIGINT/SIGTERM during shutdown is ignored.

## Test plan

- Unit tests cut a real recording off mid-way and decode the partial, stall the writer during `add` and `finish` and check the bound, and race a viewer against recording startup.
- CI runs the live verifier on an emulator for both capture sources and for the duration and size limits.
- Live, active screen, SIGKILL after 25 s: the partial held 1200 of 1362 frames and decoded cleanly. Idle screen with no viewer: only `ftyp`, no keyframe arrived.
- Live, no emulator: `/readyz` served, stop route answered 409 with the reason, exit 0 on SIGTERM.
- Live, API 35 emulator, active screen: the scrcpy server was killed on the device mid-recording. The Hub logged the capture error, a reconnecting viewer recreated the capture, and the recording finished complete with 55 frames over 28 s spanning both generations. Before this change the same kill failed the recording and the stop route answered 500, which is what a staging session hit.
- Staging, wolt project, web-preview-only Android session `01a0b361` on canary `e9db477` (2026-09-18 07:32 UTC): a 2.5 MB complete screen-recording artifact was uploaded with the full metadata; downloaded and decoded, 442 frames, 109 s, four fragments.
- serve-emu and Hub unit tests and typecheck pass.

Not verified: website playback of fragmented and partial files, long sessions.


packages/serve-emu/packages/serve-emu/src/middleware.ts
```diff
@@ -2621,6 +2639,32 @@ export function createRouter(
     return { serial, app: await getApp(serial) };
   };
 
+  const startScreenRecording = (options: RecordingOptions): Promise<void> => {
+    if (recordingStartTask) return recordingStartTask;
+    recordingStartTask = (async () => {
+      if (stopped) throw new Error("serve-emu router is stopped");
+      if (apps.size || pending.size) throw new Error("Screen recording must start before device capture.");
```
krystofwoldrich-agent, line 2654: **[suggestion]** This precondition is checked once, then `resolveSerial` and `ScreenRecording.create` are awaited before `screenRecording` is set and `ensure` runs. `createApp` reads `screenRecording?.serial === serial` at creation time only, so a viewer that creates the app inside that window gets an app without the recorder, and the start fails 15 s later with the first-frame timeout message. The CLI is safe (recording starts before `listen`), but this is a public router API. Consider setting `screenRecording` before the awaits, or re-checking `apps`/`pending` after them with a clear error, and stating the ordering on `AppOptions.screenRecording`.

packages/serve-emu/packages/serve-emu/src/screen-recording.ts
```diff
@@ -0,0 +1,373 @@
...
+    const durationUs = this.#clock.monotonicUs() - first.monotonicUs;
+    this.#enqueue(
+      this.#pending,
+      durationUs > this.#pending.timeUs ? durationUs - this.#pending.timeUs : 1n,
+    );
+    this.#pending = null;
+    try {
+      this.#throwIfFailed();
+      this.#status = { status: "finalizing" };
+      await this.#writes;
```
krystofwoldrich-agent: **[suggestion]** `await this.#writes` and `await this.#writer.finish()` have no deadline. A stalled writer keeps `finish()` pending forever; the CLI then exits 1 via its 60 s timer without writing a `failed` manifest, and the stop endpoint hangs until the client gives up. Racing finalization against a timer and routing the timeout through `fail()` would keep the failure recorded on disk and bound embedded-host shutdown too (the #114 description lists this as a known concern).

packages/expo-device-hub/src/server/index.ts
```diff
@@ -93,6 +95,22 @@ function jsonResponse(body: unknown, status = 200): Response {
 export default async function handler(request: Request): Promise<Response | null> {
   const { pathname, searchParams } = new URL(request.url);
 
+  if (pathname === '/_eas/android-recording/stop') {
```
krystofwoldrich-agent: **[suggestion]** Two small things. Other EAS routes (`/readyz`, `/metrics`) live in `eas-endpoints.ts`; this one is inline here. And on a Hub started without `--android-recording-directory`, this returns `200 {ok:true}` because `AndroidSession.finishRecording()` resolves when `startTask` is null, so a caller cannot distinguish "finalized" from "nothing to finalize". A 404/409 in that case would be more honest.

packages/expo-device-hub/README.md
```diff
@@ -60,6 +60,36 @@ the device dashboard without a running Expo project:
...
+From a built checkout, with `adb`, `ffmpeg`, and `ffprobe` on your path, run this from
+the repository root. Choose a fresh output directory and boot exactly one emulator first.
+
+```sh
+node packages/expo-device-hub/dist/server/cli.mjs \
+	--platform android \
+	--android-recording-directory /tmp/my-android-session
+```
+
+To stop, send `SIGTERM` to the Hub process, not its process group, and wait for exit.
```
krystofwoldrich-agent, line 79: **[suggestion]** `EXPO_DEVICE_HUB_RECORDING_CONTROL_TOKEN` and `POST /_eas/android-recording/stop` are only visible in `verify-android-recording.ts` and the source. Since the EAS worker depends on this contract, a short paragraph here would make it discoverable.

packages/serve-emu/packages/serve-emu/src/middleware.ts
```diff
@@ -502,6 +506,7 @@ async function createAppInternal(
     location: deviceState.lastLocation,
     route: routePlayback.snapshot(),
     session: sessionRecorder.snapshot(),
+    screenRecording: opts.screenRecording?.snapshot() ?? null,
```
krystofwoldrich-agent, line 513: **[nit]** In the `complete` state `snapshot()` includes `result.directory`, an absolute host path, on the unauthenticated `/health` route. Viewers do not need it; a reduced status without `result` would avoid exposing the path.

krystofwoldrich-agent (commented):

## Review: Add single-device Android session recording

**Stack:** PR 1/2. #114 (recording limits) is based on this branch.
**Verdict:** APPROVE (posted as a comment so the automated account does not count as a formal approval)
**Findings:** 0 critical · 0 design · 4 suggestion · 1 nit

This adds a `ScreenRecording` muxer fed from the canonical capture loop, a router-level `startScreenRecording` / `finishScreenRecording` pair that pins the capture source and encoder settings while recording, an `AndroidSession` owner in the Hub that publishes `recordings.json` only after a successful finish, an authenticated stop endpoint, and a live emulator verifier wired into CI.

The lifecycle design holds up. Finish and shutdown are idempotent and shared. Failure is fail-closed and leaves a `failed` manifest. `stop()` sets `status = "stopped"` directly and the frame loop breaks on `stopRequested` before it can reach `markTerminal`, so a normal shutdown cannot fail an in-flight finalization. Publishing happens strictly before `stopAll()`, so the result list is never written from a stale or partial state.

Verified locally at the stack head: 53 serve-emu tests (screen-recording, middleware-router), 27 Hub tests (android-session, cli-options), and the serve-emu typecheck all pass. Both new `Android recording (grpc-screenshot|scrcpy)` CI jobs are green on this head.

Nothing blocks merging. The items below are about the embedded-host contract, endpoint placement, documentation, and one unbounded wait that the #114 description already acknowledges.

### Findings

#### [suggestion] Recorder is attached only at app creation, and the precondition is checked before two awaits
`packages/serve-emu/packages/serve-emu/src/middleware.ts:2646`
`startScreenRecording` checks `apps.size || pending.size` once, then awaits `resolveSerial` and `ScreenRecording.create` before setting `screenRecording` and calling `ensure`. `createApp` reads `screenRecording?.serial === serial` at creation time only. If a viewer creates the app for that serial inside that window, the app never gets the recorder, and the start fails 15 s later with "Timed out waiting for the first recording frame", which points at the wrong cause. The CLI is safe because it starts recording before `listen`, but the router is a public API and the comment on `AppOptions.screenRecording` does not state this ordering contract. Setting `screenRecording` (or reserving the serial) before the awaits, or re-checking `apps`/`pending` after them and failing fast with a clear message, would close the gap.

#### [suggestion] Finalization has no upper bound
`packages/serve-emu/packages/serve-emu/src/screen-recording.ts:301`
`#finish` awaits the write chain and `writer.finish()` with no deadline. A stalled `mediabunny` write or a hung disk keeps `finishRecording()` pending forever. The CLI's 60 s deadline then calls `process.exit(1)` without a `failed` manifest, and a caller of `/_eas/android-recording/stop` waits until its own client timeout. The #114 description lists this as a known non-blocking concern. Racing the finalize against a timer inside `ScreenRecording` and routing the timeout through `fail()` would keep the failure on disk and make shutdown deterministic for embedded hosts too.

#### [suggestion] Stop endpoint placement and its response when nothing is recording
`packages/expo-device-hub/src/server/index.ts:98`
Other EAS-facing routes (`/readyz`, `/metrics`) live in `eas-endpoints.ts`; this one is inline in the main handler. It also returns `200 {ok:true}` on a Hub that was started without `--android-recording-directory`, because `AndroidSession.finishRecording()` resolves when `startTask` is null. A caller cannot tell "recording finalized" from "there was nothing to finalize". Consider a 404/409 in that case and grouping the route with the other EAS endpoints.

#### [suggestion] The control token and stop endpoint are not documented
`packages/expo-device-hub/README.md:78`
The README covers the CLI flag and `SIGTERM`, but `EXPO_DEVICE_HUB_RECORDING_CONTROL_TOKEN` and `POST /_eas/android-recording/stop` only appear in `verify-android-recording.ts` and the source. The companion EAS worker depends on this contract, so a short paragraph here (or in the package docs) would make the interface discoverable.

#### [nit] Health exposes the absolute recording directory
`packages/serve-emu/packages/serve-emu/src/middleware.ts:509`
`snapshot()` in the `complete` state includes `result.directory`, an absolute host path, on the unauthenticated `/health` route. The existing health payload already reports internal state, so this is consistent, but the path is not needed by viewers. Returning a reduced status (without `result`) from `snapshot()` would avoid it.

packages/serve-emu/packages/serve-emu/src/middleware.ts
krystofwoldrich-agent: Resolved in [dd14e17](https://github.com/expo/expo-device-hub/commit/dd14e173373e64b43b9e06057811bd8883d3190b): startup re-checks both `apps` and `pending` after serial resolution and recorder creation, before `ensure`, so a viewer winning the race gets the explicit ordering error instead of a first-frame timeout. [0c209de](https://github.com/expo/expo-device-hub/commit/0c209de8615b6144cadc853e7d4c29ab8aadab8a) also documents the ordering contract on `AppOptions.screenRecording`. Regression coverage: [0880748](https://github.com/expo/expo-device-hub/commit/08807487a4e7e8a7e66cf3f07061d44b8ad6d38d) races viewer creation against recording startup and checks the failure manifest.

Verified against current PR head `13cb402`. Source and regression coverage inspected; tests were not rerun for this follow-up.

— Codex

packages/serve-emu/packages/serve-emu/src/screen-recording.ts
krystofwoldrich-agent: Addressed in [c210c45](https://github.com/expo/expo-device-hub/commit/c210c45a4bb6d5749c533dba011432cab1368d66): finalization races a default 30-second deadline, routes timeout through `fail()`, and writes the failure manifest independently of the stalled writer. Waiting for that manifest is separately bounded at 5 seconds. [abff70e](https://github.com/expo/expo-device-hub/commit/abff70e80e0c0a9ba1c7432f7f65e3663c224ddd) applies bounded failure handling when recording has already failed too. Regression tests in [70c3b29](https://github.com/expo/expo-device-hub/commit/70c3b296b4ba61db14639e288f7cb4867b4ef0d9) and [f518c21](https://github.com/expo/expo-device-hub/commit/f518c21d80bed3348c8d1b952eb000657382e16e) cover stalled `add`, stalled `finish`, and an already-failed recording with a write in flight, including partial-file metadata.

Verified against current PR head `13cb402`. Source and regression coverage inspected; tests were not rerun for this follow-up.

— Codex

packages/expo-device-hub/src/server/index.ts
krystofwoldrich-agent: Both points are addressed in [6d7f420](https://github.com/expo/expo-device-hub/commit/6d7f42090296031a1b49e21bc1dc82678a8fa334). The stop route now lives in `eas-endpoints.ts` beside `/readyz` and `/metrics`. `AndroidSession.finishRecording()` reports whether a recording was published; no recording requested, skipped startup, or no result returns a reason, and the endpoint responds with 409 instead of 200. Session and endpoint tests cover the no-recording result and 200/409/500 responses.

Verified against current PR head `13cb402`. Source and regression coverage inspected; tests were not rerun for this follow-up.

— Codex

packages/expo-device-hub/README.md
krystofwoldrich-agent: Addressed in [6d7f420](https://github.com/expo/expo-device-hub/commit/6d7f42090296031a1b49e21bc1dc82678a8fa334). The README now documents `EXPO_DEVICE_HUB_RECORDING_CONTROL_TOKEN`, `POST /_eas/android-recording/stop`, the bearer header, success/no-recording/unauthorized responses, and recording-limit environment variables.

Verified against current PR head `13cb402`. Source and regression coverage inspected; tests were not rerun for this follow-up.

— Codex

packages/serve-emu/packages/serve-emu/src/middleware.ts
krystofwoldrich-agent: Addressed in [14ad849](https://github.com/expo/expo-device-hub/commit/14ad849c1ea3e6ba5e4ded16ae48baa584c1ceb7). The completed snapshot is now exactly `{ status: "complete" }`; the result containing the absolute directory is returned by `finish()` rather than included in the status exposed by `/health`. The recording test asserts that exact snapshot shape.

Verified against current PR head `13cb402`. Source and regression coverage inspected; tests were not rerun for this follow-up.

— Codex

krystofwoldrich (approved):

PR #113, opened 2026-09-16
Give Android recording an explicit session owner

vonovak:
## Why

Recording belongs to the device session, not to a preview connection. Its startup, publication and shutdown should have one explicit owner.

## How

- Add `AndroidSession` to own emulator selection, recording startup, result-list publication and shutdown.
- Share finish/shutdown promises so repeated calls finalize and publish once.
- Wait for startup before finishing; stop the router even when startup or finalization fails.
- Leave packet capture and MP4 muxing in serve-emu.

### Stack

Device Hub **2/3**, based on #112, followed by #114. Companion EAS lifecycle PR: https://github.com/expo/eas-cli/pull/4419. No standalone rollout or package publication.

## Test plan

Added lifecycle tests covering concurrent finish/shutdown, startup races, failures, unsupported device counts and terminal ownership. Updated the live verifier for viewer reconnects.

At the **cumulative stack heads**, 1,163 Hub/serve-emu tests and the serve-emu typecheck passed, and the local API → capture/events → website stop → upload → player/timeline flow passed. This was not a production Linux worker test or a separate test run at this intermediate commit.

Review note: a writer operation that never settles can keep direct/embedded recorder shutdown pending. CLI signal shutdown and EAS have outer forced-exit deadlines; direct startup/shutdown cancellation remains a hardening consideration. No recovery after Hub process failure or rotation is promised.

PR #114, opened 2026-09-16
Bound Android recording size and duration

vonovak:
## Why

Whole-session recording must have explicit disk, duration and queue limits, including bytes written while finalizing MP4 metadata. A Hub that is killed before it can finalize must still leave a playable file.

## How

- Add a positional file target enforcing a hard 4 GiB file extent and a 256 MiB free-space reserve by default. At the Hub's 6 Mbps default bitrate the one-hour duration limit fires first.
- Finalize automatically at the default one-hour duration limit, including an idle screen.
- Validate environment overrides and keep disk writes separate from preview packet delivery.
- Fail byte/disk exhaustion without publishing an incomplete recording; preserve diagnostic files.
- Write fragmented MP4. Each fragment closes at a keyframe and reaches the disk unbuffered, so a killed Hub leaves a `recording.mp4.partial` that plays up to the last keyframe, about 10 s of an active screen at the default keyframe interval.
- Write `session.json` at the first keyframe with `status: "recording"`, the partial filename, the first-frame wall-clock time and the frame size. Finish and fail overwrite it as before. The companion EAS change reads this manifest to upload the partial file.
- Start without recording when the booted-emulator count is not one. The Hub logs a warning and leaves an empty `recordings.json`, so a recording problem cannot stop a session from starting.
- Answer `POST /_eas/android-recording/stop` with 409 and the reason when nothing was recorded, so EAS can log why a session has no video. The route now lives in `eas-endpoints.ts` beside `/readyz` and `/metrics`.
- Parse the limit env vars at the boundary, naming the variable in the error and failing before any disk write.
- Fail recording startup fast when a viewer created the app during startup, instead of a 15 s first-frame timeout.
- Document the control token, the stop route and its response codes, and the three limit variables in the README.
- Keep the recording directory out of `/health`; the complete status carries only the state.
- Bound finalization at 30 s. A stalled writer fails the recording instead of hanging the stop route, embedded hosts or the CLI, and the failure manifest is written without waiting for the stalled writer. Every failed manifest names the partial file and its first-frame metadata, so EAS uploads the fragments that reached the disk with the reason.
- Extend unit and live checks for duration cutoff, byte limits, continued preview and a recording cut off before finish. The CI matrix runs the `duration` and `bytes` verifier modes as two more emulator jobs.

### Stack

Device Hub **2/2**, based on #112, which now includes lifecycle ownership from #113. Companion EAS hardening PR: https://github.com/expo/eas-cli/pull/4420.

## Follow-up tasks

- **iOS parity.** The iOS recorder (`record-sim` in eas-cli, single-file mode) still writes one MP4 with its index at finalize and writes `session.json` only at stop, so a crash loses the whole iOS recording. Set a fragment interval on its `AVAssetWriter`, write the manifest at the first accepted frame, and rebuild the packaged binary. The EAS recovery scan can then serve the iOS finish step too.
- **Keyframe cadence.** The crash-loss bound is one keyframe interval. The gRPC encoder counts its interval in frames, so an idle screen stretches it to minutes. A timed keyframe request while recording would bound the loss in wall-clock time on both sources.
- **Byte cap.** Byte exhaustion still fails the recording, while the duration cap finalizes it. The fragments written before the cap are now uploaded as a partial with the reason, so the difference is the missing tail and the `failed` status.
- **Website.** Partial recordings arrive with `partial: true` in the artifact metadata. The session page does not yet show that the video was cut off.

## Test plan

At this head:

- 57 serve-emu recorder and router tests and 40 Hub session, CLI, EAS-endpoint and limit-parsing tests passed. Two tests stall the writer during `add` and during `finish` and check that `finish()` rejects within the bound and the manifest lists the partial file. serve-emu typecheck and `git diff --check` passed.
- A new unit test cuts a real recording off after four frames and decodes the `.partial` with ffprobe and ffmpeg. `session.json` at that point carries `status: "recording"` and the first-frame metadata.
- Live on an `eas_model_small` emulator with a swipe loop keeping the screen active: the Hub was killed with SIGKILL after 25 s of gRPC capture. The `.partial` held 1200 of 1362 frames and 22.4 s, and ffmpeg decoded it without errors. The same test on an idle screen with no viewer left only the `ftyp` box, because no keyframe arrived in 25 s.
- With no emulator booted and recording requested, the built Hub served `/readyz`, logged `Android recording skipped`, answered the stop route with `409 {"error":"Android recording requires exactly one booted emulator; found 0."}`, exited 0 on SIGTERM and left `recordings.json` as `[]`.
- Both CI verifier cases passed live: scrcpy with signal stop and gRPC with endpoint stop. The verifier no longer asserts per-packet durations, which ffprobe reports as `N/A` for fragmented files; sample durations are stored in each fragment header and the total duration is still checked.
- Companion EAS head: 1485 build-tools tests and typecheck passed, including the recovery of a real partial left by a killed Hub.

Rotation and codec changes still discard the recording by design. Long-session reliability and website playback of fragmented files are not yet verified on hosted workers.


packages/serve-emu/packages/serve-emu/src/recording-file-target.ts
```diff
@@ -0,0 +1,66 @@
...
+  let checkedAt = performance.now();
+  let closeTask: Promise<void> | null = null;
+  const close = () => (closeTask ??= file.close());
+  const target = new StreamTarget(
+    new WritableStream<StreamTargetChunk>({
+      async write({ data, position }) {
+        try {
+          const end = position + data.byteLength;
+          if (!Number.isSafeInteger(end) || end > options.maxFileBytes) {
+            throw new Error("Recording exceeded its file byte limit.");
```
krystofwoldrich-agent, line 32: **[suggestion]** This throw discards the entire recording once the cap is reached, while the duration limit finalizes what exists. Since `moov` is written at the end, the target could reserve headroom and ask the recorder to finish (a `#finishAtByteLimit()` mirroring `#finishAtDurationLimit()`) before the hard cap, keeping this throw as a backstop. If discard-on-exhaustion stays by design, the EAS side should treat `failed` + `byte limit` as "too large" rather than a capture fault.

packages/expo-device-hub/src/server/serve-emu.ts
```diff
@@ -28,7 +28,18 @@ export const emuCameraFeeds: EmulatorCameraFeeds = {
...
+    maxFileBytes: recordingLimitFromEnv('EXPO_DEVICE_HUB_RECORDING_MAX_BYTES'),
+    maxDurationMs: recordingLimitFromEnv('EXPO_DEVICE_HUB_RECORDING_MAX_DURATION_MS'),
+    minFreeBytes: recordingLimitFromEnv('EXPO_DEVICE_HUB_RECORDING_MIN_FREE_BYTES'),
+  });
+
+function recordingLimitFromEnv(name: string): number | undefined {
+  const value = process.env[name];
+  if (value === undefined) return undefined;
+  // ScreenRecording.create validates the public options, including environment overrides.
+  return value.trim() === '' ? NaN : Number(value);
```
krystofwoldrich-agent: **[suggestion]** Mapping `""` to `NaN` makes `ScreenRecording.create` reject it, but the error is `Invalid recording limit maxFileBytes.` and never names `EXPO_DEVICE_HUB_RECORDING_MAX_BYTES`. On a hosted worker that is the only clue in the log. Validating here with the variable name in the message, and listing the three variables in the README, would make misconfiguration diagnosable.

krystofwoldrich-agent (commented):

## Review: Bound Android recording size and duration

**Stack:** PR 2/2, based on #112.
**Verdict:** APPROVE (posted as a comment so the automated account does not count as a formal approval)
**Findings:** 0 critical · 0 design · 2 suggestion · 0 nit

This adds a positional `StreamTarget` that enforces a file byte limit and a free-space reserve on every write (including `moov` finalization), a duration limit that finalizes cleanly with `stopReason: "duration-limit"`, validated limit options, environment overrides in the Hub, and unit plus live-verifier coverage for all three limits.

The file target is correct for the positional writes `mediabunny` performs with `fastStart: false`: it tracks the extent rather than summing bytes, so header rewrites do not double-count, and the free-space probe is refreshed every 5 s and debited between probes. `createMp4Writer` closes the handle on `start`, `finish`, and `cancel` failures, and `close` is memoized. The duration path clears the timer in both `finish()` and `fail()`, and `accept()` ignores packets once the recorder is no longer active, so the preview keeps running after the cutoff.

Verified locally at this head: 53 serve-emu tests and 27 Hub tests pass, serve-emu typecheck passes. Both `Android recording` CI jobs are green.

### Findings

#### [suggestion] Byte and disk exhaustion discard the whole recording, while the duration limit finalizes
`packages/serve-emu/packages/serve-emu/src/recording-file-target.ts:32`
Hitting the 2 GiB default after, say, 50 minutes throws away everything written so far. The duration limit shows the alternative: stop accepting frames and finalize what exists. Because `moov` is written at the end, the target could reserve headroom (a few MiB plus a per-frame estimate) and trigger a `#finishAtByteLimit()` on the recorder before the hard cap, keeping the hard throw only as a backstop. The PR description says discard-on-exhaustion is by design; if that stays, the manifest `error` is the only signal, so the EAS side should treat `failed` + `byte limit` as "recording existed but was too large" rather than a capture fault.

#### [suggestion] Environment override errors do not name the variable, and the variables are undocumented
`packages/expo-device-hub/src/server/serve-emu.ts:41`
`recordingLimitFromEnv` maps `""` to `NaN` so that `ScreenRecording.create` rejects it, which works, but the resulting message is `Invalid recording limit maxFileBytes.` and never mentions `EXPO_DEVICE_HUB_RECORDING_MAX_BYTES`. On a hosted worker that is the only clue in the log. Validating here with the variable name in the error, and adding the three variables to the README section from #112, would make misconfiguration diagnosable.

vonovak:
Folded into #112. Fast-forwarding the #112 branch to this branch's head made GitHub mark this PR as merged; the diff and commits now live in #112. Both suggestions here are addressed on that head: the byte-cap default is 4 GiB so the duration limit fires first (5c15961), and limit env vars are validated at the boundary with the variable name in the error (6d7f420). The byte-exhaustion policy is listed as a follow-up in #112.

PR #117, opened 2026-09-17
Show the Android emulator's runtime GPU backend

krystofwoldrich-agent:
Emulators started with `-gpu auto` can select a software renderer without making that visible in Hub. Add a read-only **GPU backend** row to Device options that displays the renderer reported by `adb shell dumpsys SurfaceFlinger`, with a short name, graphics details, and the full renderer in a tooltip.

Android utils bounds the ADB read to three seconds and returns `null` on failure, timeout, or missing renderer information. Hub reads this through a dedicated backend endpoint for the selected emulator and displays **Unknown** when unavailable. There are no launch-flag or AVD-configuration fallbacks; the row is hidden for iOS and physical Android devices.

Validation: Android utils and Hub/components tests pass, including timeout/failure handling and UI visibility. Utils/components lint and typechecks, and Hub web/server builds pass. The full Hub typecheck reports existing errors in the unchanged `serve-emu-options.test.ts`. Verified the real backend and captured the live UI with agent-browser on `emulator-5554`, reporting **SwiftShader · Software · ANGLE / Vulkan**; no browser page errors were reported.

Live Android emulator with the GPU backend row showing SwiftShader and Software · ANGLE / Vulkan


packages/@expo/hub-android-utils/src/gpu-info.ts
```diff
@@ -0,0 +1,112 @@
...
+  }: {
+    timeoutMs?: number;
+    signal?: AbortSignal;
+    /** Override SDK resolution, e.g. for a custom adb installation. */
+    adbPath?: string;
+  } = {},
+): Promise<AndroidUtilsResult<AndroidGpuInfo | null>> {
+  try {
+    const { stdout } = await execFileAsync(
+      adbPath ?? resolveAdbPath(process.env, homedir()),
```
krystofwoldrich-agent, line 101: **[suggestion]** Every other adb invocation in this package is a small runner in `adb.ts` that takes `adbPath` first (`runAdbGetprop`, `runAdbEmuAvdName`, `runAdbEmuKill`), and the public functions resolve the path once via `resolveAdbPath`. This inlines the `execFile` call and adds an `adbPath` option override, a new pattern for the public API. A `runAdbDumpsysSurfaceFlinger(adbPath, serial, { timeoutMs, signal })` in `adb.ts` would keep parsing here, keep the test seam, and match the existing shape.

packages/expo-device-hub/src/dashboard/useDeviceGpu.ts
```diff
@@ -0,0 +1,45 @@
+import { type AndroidGpuInfo } from '@expo/hub-android-utils';
```
krystofwoldrich-agent, line 1: **[suggestion]** This is the first file under `src/dashboard` that imports from `@expo/hub-android-utils`, a package that pulls in `node:child_process`. The `type` modifier erases it today, but it couples the Metro web bundle to a server-side package, and a later non-type import would break the client build. The PR already defines the structurally identical `DeviceGpuInfo` in `DeviceOptionsSection.tsx` and exports it from `@expo/hub-components`. Importing that here removes the duplicate type and keeps the client/server boundary.

packages/expo-device-hub/src/dashboard/useDeviceGpu.ts
```diff
@@ -0,0 +1,45 @@
...
+        // An unavailable renderer is displayed as Unknown.
+      } finally {
+        clearTimeout(timeout);
+      }
+    })();
+    return () => {
+      controller.abort();
+      clearTimeout(timeout);
+    };
+  }, [serial]);
```
krystofwoldrich-agent, line 42: **[suggestion]** The effect runs once per `serial`. An emulator enters the booted list as soon as `adb devices` reports state `device`, which is early in boot, before SurfaceFlinger has necessarily initialized its GL context. If the first `dumpsys` read fails or lacks a `GLES:` line at that point, the row shows **Unknown** until the user selects another device and comes back. Consider retrying a `null` result once after a short delay, or re-reading when the device client reconnects.

packages/@expo/hub-android-utils/src/gpu-info.ts
```diff
@@ -0,0 +1,112 @@
...
+  fields.push(value.slice(start).trim());
+  return fields;
+}
+
+/** Parse the GLES vendor/renderer/version line, preserving unfamiliar renderers. */
+export function parseGpuInfo(dump: string): AndroidGpuInfo | null {
+  const line = /^[\t ]*GLES:[\t ]*([^\r\n]*)/m.exec(dump)?.[1];
+  if (!line) return null;
+  const fields = splitFields(line);
+  if (fields.length !== 3 || !fields[1]) return null;
```
krystofwoldrich-agent, line 40: **[nit]** `fields.length !== 3` rejects the whole line when the vendor or version part contains a top-level comma. The renderer is always the middle part, so slicing between the first and last top-level comma would still show the raw renderer in those cases instead of **Unknown**.

krystofwoldrich-agent (commented):

## Review: Show the Android emulator's runtime GPU backend

**Verdict:** APPROVE (posted as a comment: automated account, own PR)
**Findings:** 0 critical · 0 design · 3 suggestion · 2 nit

The change adds a read-only **GPU backend** row for Android emulators. A new `readGpuInfo` in `@expo/hub-android-utils` runs `adb -s <serial> shell dumpsys SurfaceFlinger` with a 3 s timeout and parses the `GLES:` line. The Hub server exposes it at `GET /api/devices/gpu?serial=emulator-N`, validates the serial with a strict regex, and never shells out for other methods or serials. The dashboard reads it once per selected emulator and shows **Unknown** when the read fails. The boundaries are sound: `execFile` with an argument array, a bounded child process, and a route that returns 200 with `gpu: null` on failure so the UI degrades quietly.

The suggestions below are about where the adb runner lives, a type import that reaches from the web bundle into a Node-only package, and the one-shot read that can leave **Unknown** on a still-booting emulator.

### Also worth noting

- **No changeset.** Recent user-visible features on `expo-device-hub` shipped with a `.changeset/*.md` (for example the Permissions section, activity metrics, and app icon PRs). This PR adds a visible row without one, so it will not appear in the changelog or bump the package.
- **CI has not run.** The workflow run for this head is in `action_required` state, so no checks are reported on the PR yet. A maintainer needs to approve the run.

### Local verification

- `@expo/hub-android-utils`: 143 tests pass (18 files, including the 17 new `gpu-info` cases), `tsc --noEmit` and `oxlint` clean.
- `@expo/hub-components`: the new `device-gpu.test.tsx` passes (5 tests); `oxlint` clean on the changed files.
- `expo-device-hub`: the new `device-gpu.test.ts` passes (5 tests), and the new files typecheck.
- I could not install dependencies in this environment, so I ran the suites against the existing local install. Tests that render the real Radix select (LocationSection and inspectorSections) failed only because that package is missing locally; the failures are unrelated to this PR.

### Findings

#### [suggestion] Put the adb runner beside the other `adb -s <serial>` helpers
`packages/@expo/hub-android-utils/src/gpu-info.ts:101`
Every other adb invocation in this package is a small runner in `adb.ts` that takes `adbPath` as its first argument (`runAdbGetprop`, `runAdbEmuAvdName`, `runAdbEmuKill`), and the public functions resolve the path once via `resolveAdbPath`. `readGpuInfo` inlines the `execFile` call and introduces an `adbPath` option override, which is a new pattern for the public API. A `runAdbDumpsysSurfaceFlinger(adbPath, serial, { timeoutMs, signal })` in `adb.ts` would keep the parsing in `gpu-info.ts`, keep the test seam, and match the existing shape.

#### [suggestion] Take the row type from `@expo/hub-components` instead of the Node-only utils package
`packages/expo-device-hub/src/dashboard/useDeviceGpu.ts:1`
This is the first file under `src/dashboard` that imports from `@expo/hub-android-utils`, a package that pulls in `node:child_process`. The `type` modifier erases it today, so the web build passes, but it couples the Metro web bundle to a server-side package, and a later non-type import would break the client build. The PR already defines the structurally identical `DeviceGpuInfo` in `DeviceOptionsSection.tsx` and exports it from `@expo/hub-components`. Importing that here removes the duplicate type and keeps the client/server boundary that the rest of `src/dashboard` follows.

#### [suggestion] A single read can leave **Unknown** on a still-booting emulator
`packages/expo-device-hub/src/dashboard/useDeviceGpu.ts:42`
The effect runs once per `serial`. An emulator enters the booted list as soon as `adb devices` reports state `device`, which happens early in boot, before SurfaceFlinger has necessarily initialized its GL context. If the first `dumpsys` read fails or has no `GLES:` line at that point, the row shows **Unknown** until the user selects another device and comes back. Consider retrying a `null` result once after a short delay, or re-reading when the device client reconnects.

#### [nit] Strict three-field parse turns any unexpected line shape into **Unknown**
`packages/@expo/hub-android-utils/src/gpu-info.ts:40`
`fields.length !== 3` rejects the whole line when the vendor or version part contains a top-level comma. The renderer is always the middle part, so slicing between the first and last top-level comma would still show the raw renderer in those cases instead of hiding it.

#### [nit] Share one JSON response helper with the other API routes
`packages/expo-device-hub/src/server/device-gpu.ts:8`
The other routes in `server/index.ts` answer through `jsonResponse`, which sets `Content-Type: application/json; charset=utf-8` and `Cache-Control: no-store`. This handler builds its own headers with `Response.json`, so the charset differs. Exporting the helper (or moving it to a shared module) would keep the API responses uniform.

packages/expo-device-hub/src/server/device-gpu.ts
```diff
@@ -0,0 +1,18 @@
+import { readGpuInfo } from '@expo/hub-android-utils';
+
+/** Selected-emulator metadata is read on demand, outside device discovery. */
+export async function handleDeviceGpuRequest(
+  request: Request,
+  readGpu = readGpuInfo,
+): Promise<Response> {
+  const headers = { 'Cache-Control': 'no-store' };
```
krystofwoldrich-agent, line 8: **[nit]** The other routes in `server/index.ts` answer through `jsonResponse`, which sets `Content-Type: application/json; charset=utf-8` plus `Cache-Control: no-store`. This handler builds its own headers with `Response.json`, so the charset differs. Exporting that helper (or moving it to a shared module) would keep API responses uniform.

PR #120, opened 2026-09-21
chore(hub-client): prepare @expo/hub-client for publishing to npm

krystofwoldrich-agent:
## Summary

Prepares `@expo/hub-client` for publishing to npm as a standalone package, adds a package README, and updates the release docs. The package will mostly be consumed internally, but it is public, so the tarball and manifest should look like a normal npm package.

## Review: what was ready

- **Source and API.** The public entry point re-exports the `DeviceClient` contract, the two platform hooks, `useActiveDeviceClient`, `DeviceScreen`, and the orientation helpers. Every type carries a doc comment.
- **Build.** swc emits ESM (`.mjs`) plus `tsc` declaration files with declaration maps; source maps resolve to the shipped `src/`.
- **Manifest shape.** `type: module`, `exports` with `types`/`import`/`default`, `sideEffects: false`, and `react` as the only runtime dependency (declared as a peer).
- **Tests and checks.** 293 tests, typecheck, and lint pass; CI already runs them for this package.

## Review: what needed changes (done in this PR)

- **`private: true` blocked publishing.** Removed. `changeset publish` and the release scripts skip private packages, so nothing in the release flow reached this package before.
- **Missing npm metadata.** Added `repository` (with `directory`), `homepage`, `bugs`, `author`, and `publishConfig.access: public`. The `repository.url` matches the workflow repository, which npm requires for a provenance-signed publish.
- **Tarball contents.** `npm pack` shipped all 42 test files under `src/__tests__` and had no README or LICENSE. Tests are now excluded via `!src/__tests__` in `files`; README and an MIT LICENSE are included. Before: 318 files, 1.7 MB unpacked. After: 278 files, 1.5 MB unpacked.
- **ESM output was not resolvable by Node.** The emitted `.mjs` files imported `./types` without an extension, so `import('@expo/hub-client')` failed under Node's ESM resolver. swc now runs with `module.resolveFully` and emits `./types.mjs`. Bundler consumers (Metro, webpack, Vite) are unchanged.
- **Peer range.** `react: "*"` is now `react: ">=18.0.0"`. The code uses only long-standing hooks and `react/jsx-runtime`.
- **Changeset.** Added a `minor` changeset for `@expo/hub-client` so the next real release versions and publishes it (`changeset status` shows `0.2.0 → 0.3.0`).
- **Docs.** New `packages/@expo/hub-client/README.md` with the two main flows: rendering a live device screen and calling device controls. `RELEASING.md` and `.changeset/README.md` now describe two public packages, the canary release that runs on every push to `main`, and the one-time npm setup a new public package needs. The root README links to the package README.

## Release workflow review

`release.yml` and the scripts under `scripts/` work for this package once it is public: `getPublicPackages` picks it up, `changeset version` bumps it from its changeset, `apply-canary-version.ts` and `ignore-packages.ts` handle it, and `create-github-releases.ts` packs and attaches it. Two things need attention **before this merges**:

1. **npm Trusted Publisher for `@expo/hub-client`.** The workflow publishes with OIDC only. The package must have a GitHub Actions trusted publisher configured on npmjs.com (repository `expo/expo-device-hub`, workflow `release.yml`, environment `npm-publish`). Without it, the canary publish that runs on every push to `main` will fail for this package as soon as this PR lands. `changeset publish` attempts every package, so the `expo-device-hub` canary would still publish, but the workflow run would be red.
2. **Existing placeholder on npm.** `@expo/hub-client@1.0.0` already exists on npm (published 2026-09-16, 1 file, 315 bytes, empty description). The next release from this repo is `0.3.0`, which is lower. `npm publish` moves `latest` to `0.3.0`, so plain installs work, but the `1.0.0` version stays visible and tools that pick the highest version will offer it. Options:
   - Keep `0.x` and deprecate the placeholder: `npm deprecate @expo/hub-client@1.0.0 "Placeholder release, use 0.x"`. This is the smaller change and keeps the pre-1.0 signal for an API that still moves often.
   - Or set the package version to `1.0.0` locally so the changeset produces `1.1.0`. This skips the conflict but declares a stable API.

## Known limitation (not changed here)

TypeScript consumers with `moduleResolution: node16` or `nodenext` cannot resolve the re-exported types, because the emitted `.d.ts` files import `./types` without an extension. `bundler` and `node10` resolution work. Fixing it means either adding `.js` extensions to every relative import in `src/` and emitting `.js` files, or bundling the declarations into one file. Both touch the whole package, so they are left for a follow-up if a NodeNext consumer appears.

## Verification

- `bun test` in the package: 293 pass, 0 fail.
- `bun run typecheck` and `bun run lint` in the package; root `bun run lint`; turbo build of `@expo/hub-client` and `@expo/hub-components` and typecheck of `@expo/hub-components`.
- `bun install --frozen-lockfile` passes with the updated lockfile.
- `npm pack` into a throwaway project with React 19 and TypeScript 5.9: `tsc` with `moduleResolution: Bundler` passes and `node --input-type=module -e "import('@expo/hub-client')"` resolves all 10 exports.
- `bunx changeset status` lists `@expo/hub-client 0.3.0` under minor bumps.

The `bun.lock` diff also refreshes the recorded `expo-device-hub` version (`0.9.0 → 0.10.1`), which the last release commit had left stale.


.changeset/publish-hub-client.md
```diff
@@ -0,0 +1,5 @@
+---
+'@expo/hub-client': minor
+---
+
+Publish `@expo/hub-client` to npm as a standalone package. The package is no longer private, ships a README and LICENSE, leaves its tests out of the tarball, and emits ESM with explicit `.mjs` import extensions so Node can resolve it.
```
krystofwoldrich, line 5:
````suggestion
Initial publish
````

